PR #9, opened 2021-05-20
manifest: default `ami_name` to match capa default

jkoelker:
Make the entire `ami_name` a user variable, and default it to the prefix
`capa-ami-` which will allow it to be discovered by capa by default.


joejulian (approved):

jkoelker:
Looks like the `clean_resource_name` function is not valid for the the variable interpolation, moved it back into the `builders[].ami_name`.

jkoelker:
After discussing, we've decided to keep it prefixed with `konvoy-ami` for the time being and instead update konvoy to support setting the capa prefix.

PR #20, opened 2021-06-15
fix: override flags not passed on generate command

mhrabovcin:
The `generate` command was ignoring `overrides` CLI flag.

```
./bin/konvoy-images generate [path] --overrides [override-path]
```


jr0d (approved):
ty!

PR #29, opened 2021-07-01
[WIP] ansible molecule testing

fatz:
Have Ansible modules being tested on their own.

Goals:
- define tests for a successful run
- define supported OS versions as platforms
- define special case scenarios like GPU and test them with GPU instance types

Test driven Ansible:
Adding new OS distributions and versions to platforms and then add modules to make the tests work


fatz:
#34 integrates this

PR #33, opened 2021-07-06
Disable automatic updates for Flatcar

jongiddy:
To ensure the Nvidia driver continues to work, we need to be conservative with kernel upgrades. Flatcar in particular upgrades automatically. Disable upgrades (and their reboots) on Flatcar to provide a similar model to other OS's.

Since the OS does not upgrade, encode the OS version in the AMI name. This requires retrieval of the AMI data before running the AMI build. This is done by running a small HCL build (in `hack/flatcar-ami.pkr.hcl`) that fetches the latest Flatcar AMI and creates an override file to set the `distribution_version`.

https://jira.d2iq.com/browse/D2IQ-76956

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ export DOCKER_DEVKIT_AWS_ARGS ?= \
 	--env AWS_PROFILE \
 	--env AWS_SECRET_ACCESS_KEY \
 	--env AWS_SESSION_TOKEN \
-	--env AWS_REGION \
+	--env AWS_DEFAULT_REGION \
 	--volume "$(HOME)/.aws":"/home/$(USER_NAME)/.aws"
 
 ifneq ($(wildcard $(DOCKER_SOCKET)),)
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -136,7 +136,8 @@ centos8-nvidia: ## Build Centos 8 image
 .PHONY: flatcar
 flatcar: build
 flatcar: ## Build flatcar image
-	./bin/konvoy-image build images/ami/flatcar.yaml
+	AWS_DEFAULT_REGION=$(AWS_REGION) ./hack/fetch-flatcar-ami.sh
```
mhrabovcin, line 139: Should we make this a separate make target that creates `flatcar-version.yaml`.

jongiddy: On one hand, that saves fetching the data repeatedly. On the other hand, it may mean we attempt to build an older version when a new version becomes available. But it would make the e2e test cleaner, since I just added the same command there (although the e2e tests seems to pass without it).

Makefile
```diff
@@ -301,7 +303,7 @@ ifeq ($(CI), true)
 export DOCKER_DEVKIT_AWS_ARGS := --env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY
 endif
 
-ci.e2e.build.all:
+ci.e2e.build.all: flatcar-version.yaml
```
mhrabovcin, line 306: Shouldn't be the `flatcar-version.yaml` executed in the devkit container?

jongiddy: True. Is there a reason this doesn't just run:
```shell
WHAT="make flatcar" make devkit.run
```

Makefile
jongiddy: Changed to use a separate targer

Makefile
jongiddy: Fixed

fatz (approved):

mhrabovcin (approved):

PR #34, opened 2021-07-07
Suse initial support

fatz:
Added Suse support and tests:

- [x]  do central decision which task file to use instead of deciding on every resource
- [x] add suse repositories
- [x] add suse package installation
- [x] add suse version lock
- [x] add suse cloud tools install
- [x] add nvidia support

took the DC/OS gpu module (https://github.com/dcos/dcos-ansible/tree/master/molecule/ec2_gpu) and added suse to it.

Also changed `gpu.type` fact to `gpu.types` so a user could potentially install more than just nvidia drivers.

GPU support:
- [x] unify version variables for cuda and cuda driver


mhrabovcin:
```
fatal: [default]: FAILED! => {"msg": "The conditional check ''nvidia' in gpu.types' failed. The error was: error while evaluating conditional ('nvidia' in gpu.types):
```

centos7-nvidia build is failing

ansible/roles/packages/tasks/nvidia.yaml
```diff
@@ -1,84 +0,0 @@
...
-        - ansible_os_family == 'RedHat'
-        - ansible_distribution_major_version|int == 7
-    - name: Install Nvidia driver (Centos/RHEL 8)
-      dnf:
-        name: "@{{ nvidia_driver_dnf_module_name }}:\
-          {{ nvidia_driver_dnf_module_version }}"
-        state: present
-      when:
-        - ansible_os_family == 'RedHat'
-        - ansible_distribution_major_version | int >= 8
```
mhrabovcin, line 84: @fatz are you removing centos8 / rhel8 nvidia support on purpose? I don't see this being moved to the new `gpu` module.

fatz: I'll have a look.

ansible/requirements.txt
```diff
@@ -0,0 +1,8 @@
+molecule
+molecule-ec2
+ansible-lint
+boto
+boto3
+ansible==2.9.20
```
mhrabovcin, line 6: We use different ansible in `konvoy-image-builder` container https://github.com/mesosphere/konvoy-image-builder/blob/eb4ac11723e7ca19ed7392d923299bd7fa5177d7/Dockerfile#L9

fatz: thanks a lot yeah makes sense I'll change that and run the tests again

ansible/group_vars/all/system.yaml
```diff
@@ -13,7 +16,12 @@ docker_rpm_repository_url_upstream_mirror: "https://packages.d2iq.com/download.d
 docker_rpm_repository_url: "https://packages.d2iq.com/konvoy/stable/linux/centos/{{ ansible_distribution_major_version|int }}/x86_64"
 docker_rpm_gpg_key_url: "{{ docker_rpm_repository_url }}/gpg.asc"
 
+
+docker_rpm_suse_repository_url: "https://packages.d2iq.com/konvoy/stable/linux/opensuse/leap/{{ ansible_distribution_major_version|int }}/x86_64"
+docker_rpm_suse_gpg_key_url: "{{ docker_rpm_suse_repository_url }}/gpg.asc"
+
 # nvidia package
+nvidia_cuda_version: "460"
```
mhrabovcin, line 24: Is the `cuda` word a synonym for `driver`?

fatz: hmm its actually the framework... they sometimes mix it but just saw not in case of the driver packge... let me see if i can rename the vars

ansible/molecule/default/tests/test_default.py
```diff
@@ -0,0 +1,27 @@
+import os
```
mhrabovcin, line 1: The existing `konvoy-image-builder` is using `goss` packer plugin https://github.com/YaleUniversity/packer-provisioner-goss as a provisioning step that should test provisioned packages. Is this is a different level test? Should these 2 be unified?

ansible/molecule/default/molecule.yml
```diff
@@ -0,0 +1,52 @@
...
+scenario:
+  name: default
+driver:
+  name: ec2
+platforms:
+  - name: konvoyimage-centos7.9-${USER:-ci}-${HOSTNAME:-local}
+    image_search_name: "CentOS 7.9.2009 x86_64"
+    image_search_owner: "125523088429"
+    region: us-east-1
+    instance_type: m4.xlarge
```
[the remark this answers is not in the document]
fatz: it shouldnt make to much of a difference but from my experience using t instances with test runs adds quite some waiting time as they could get capped especially when building nvidia driver. 

But maybe large works aswell. I'll try that

but good point.

ansible/molecule/ec2_gpu/molecule.yml
```diff
@@ -0,0 +1,69 @@
...
+scenario:
+  name: ec2_gpu
+driver:
+  name: ec2
+platforms:
+  - name: konvoyimage-centos7.9-${USER:-ci}-${HOSTNAME:-local}
+    image_search_name: "CentOS 7.9.2009 x86_64"
+    image_search_owner: "125523088429"
+    region: us-east-1
+    instance_type: p2.xlarge
```
mhrabovcin, line 11: Do we need to execute builds on `p2.xlarge` instance types?

ansible/molecule/default/tests/test_default.py
fatz: not sure if they should be unified. Theoretically they could be. 

With molecule you do not want to execute gpu tests if you work on something different and vice versa. Thats why we have two different scenarios.

I do not know goss it seems to be yaml based testing. I'm not sure how easy it is to adapt the same functionality. What was the reason to not use code based testing?

ansible/molecule/default/molecule.yml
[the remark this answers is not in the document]
mhrabovcin: I think it would be better to run tests on the same instance types as where the end-users would execute the builds.

ansible/molecule/ec2_gpu/molecule.yml
fatz: yup with that the tests ensures the kernel successfully loaded the module as nvidia-smi is being used

ansible/roles/gpu/tasks/nvidia-gpu-CentOS7.yaml
```diff
@@ -0,0 +1,91 @@
...
+        baseurl: "{{ vault_repo_baseurl }}"
+        enabled: false
+    - name: ThisVault - Install Kernel Headers for Current Kernel
+      yum:
+        enablerepo: ThisVault
+        name: "kernel-headers-{{ hostvars[inventory_hostname].ansible_kernel | regex_search('\\d+\\.\\d+\\.\\d+-\\d+') }}.el{{ hostvars[inventory_hostname].ansible_distribution_major_version }}"
+    - name: ThisVault - Install Kernel Devel for Current Kernel
+      yum:
+        enablerepo: ThisVault
+        name: "kernel-devel-{{ hostvars[inventory_hostname].ansible_kernel | regex_search('\\d+\\.\\d+\\.\\d+-\\d+') }}.el{{ hostvars[inventory_hostname].ansible_distribution_major_version }}"
```
mhrabovcin, line 43: Is `ThisVault` some commonly used name?

fatz: there are multiple vaults for a single major version. So every centos minor version has its own vault. The repo is disabled by default so its should not interfere with something else.

Would `ThisCentOSVault` be a better matching name? I tried to avoid variable names for this making it a bit easier to debug

mhrabovcin: I am ok with keeping the name. I was just curious.

fatz:
I found another thing. I forgot the container runtime. 

I'll see if i can add a test for that

fatz:
about GPU

i realized that we're statically set config when nvidia is set. That means users must have different images for gpu and non gpu support.

We should probably find a way to do this on boot time instead. This would allow users using a single image for gpu and non-gpu instances

images/ami/sles-15.yaml
```diff
@@ -0,0 +1,15 @@
...
+  ami_filter_name: "suse-sles-15-sp3-v*"
+  ami_filter_owners: "amazon"
+  distribution: "SLES"
+  distribution_version: "15"
+  source_ami: ""
+  ssh_username: "ec2-user"
+  root_device_name: "/dev/sda1"
+  volume_size: "15"
+
+build_name: "centos-8"
```
jongiddy, line 13:
```suggestion
build_name: "sles-15"
```

Makefile
```diff
@@ -306,16 +308,21 @@ endif
...
-	WHAT="./bin/konvoy-image build images/ami/centos-8.yaml -v 6" make devkit.run
+	WHAT="./bin/konvoy-image build images/ami/sles-15.yaml --overrides overrides/nvidia.yaml -v ${VERBOSITY}" make devkit.run
 	make docker.clean-latest-ami
-	WHAT="./bin/konvoy-image build images/ami/centos-7.yaml --overrides overrides/nvidia.yaml -v 6" make devkit.run
+	WHAT="./bin/konvoy-image build images/ami/centos-7.yaml --overrides overrides/nvidia.yaml -v ${VERBOSITY}" make devkit.run
 	make docker.clean-latest-ami
-	WHAT="./bin/konvoy-image build images/ami/centos-8.yaml --overrides overrides/nvidia.yaml -v 6" make devkit.run
+	WHAT="./bin/konvoy-image build images/ami/centos-8.yaml --overrides overrides/nvidia.yaml -v ${VERBOSITY}" make devkit.run
+	make docker.clean-latest-ami
+	WHAT="./bin/konvoy-image build images/ami/sles-15.yaml --overrides overrides/nvidia.yaml -v ${VERBOSITY}" make devkit.run
```
jongiddy, line 321: The first SLES build should be without the override for GPU support.

The Makefile also includes targets for `centos7` and `centos7-nvidia`. Can you please add similar for Suse?

jongiddy (commented):
Looking good! Just some minor fixes needed.

ansible/molecule/default/tests/test_default.py
jr0d: goss is used to validate that a device is configured properly for a successful bootstrap. It is run after ansible provisioning to validate that the ansible was successful; but it is also capable of validating a device which was provisioned outside of KiB.

Makefile
fatz: done ty

ansible/roles/packages/tasks/install-flatcar.yaml
```diff
@@ -0,0 +1,120 @@
...
+    dest: /opt
+    extra_opts:
+      - --no-overwrite-dir
+      - --strip-components=2
+      - --wildcards
+      - '*containerd-shim-runc-v*'
+
+- name: Create symlinks for containerd
+  file:
+    src: /usr/bin/{{ item }}
```
jr0d, line 54: flatcar instances typically are only writable under /opt and /etc

jr0d (changes requested):
All in all, this looks great. For flatcar, I think we need to remove the symlink under /usr/bin for containerd.

ansible/roles/packages/tasks/install-flatcar.yaml
fatz: good point. think it it runs through probably because links are already there

fatz: i think this can be handled in the seperate flatcar PR. This just took over what was already in image builder but in separated files.

ansible/molecule/default/tests/test_default.py
fatz: for now I would not want to change anything. But my suggestion for the future would be that different molecule scenarios have different tests and the EE test merges them.

It doesn't really matter which technology is used.

With a second though I currently prefer the way its currently as we have two different tests ensuring things. The molecule tests are currently more for local development.

ansible/group_vars/all/system.yaml
fatz: nvidia is mixing the naming by themself. They call sometimes the driver package cuda-drivers. So I think its fair to stick with this name

jongiddy:
I'm merging this. I haven't found any problems in testing, and it appears to work no worse than previously for Flatcar. https://github.com/mesosphere/konvoy-image-builder/pull/39 improves Flatcar support.

PR #45, opened 2021-07-30
fix: pause image

dkoshkin:
Bringing back the pause image that was removed in https://github.com/mesosphere/konvoy-image-builder/commit/b0a8451c00d9ac631061281c97ecb760c474dc21

TODO: need to figure out where this image version is being referenced.
* It was not in Containerd configs
* Not in the Kubelet configs
* Not in the bootstrap config


ansible/group_vars/all/defaults.yaml
```diff
@@ -6,6 +6,8 @@ kubernetes_cni_version: "0.9.1"
 etcd_version: "3.4.13-0"
 coredns_version: "1.8.0"
 pause_image_version: "3.4.1"
```
jkoelker, line 8: Should we make this a list of `pause_image_versions` and pull all the versions in the list?

ansible/roles/packages/templates/config.toml.tmpl
```diff
@@ -62,7 +62,7 @@ imports = ["/etc/containerd/config-konvoy-*.toml"]
     stream_server_port = "0"
     stream_idle_timeout = "4h0m0s"
     enable_selinux = false
-    sandbox_image = "{{ default_image_registry }}k8s.gcr.io/pause:3.1"
+    sandbox_image = "{{ default_image_registry }}k8s.gcr.io/pause:{{ pause_image_version }}"
```
jkoelker, line 65: Nice find!

jkoelker (approved):
LGTM, just a question on how often we may need to use multiple image versions and if it should then be a list, I dont' really have a good gauge for that so nothing blocking.

dkoshkin:
> LGTM, just a question on how often we may need to use multiple image versions and if it should then be a list, I dont' really have a good gauge for that so nothing blocking.

We will need to spend some time figuring out if this is a misconfiguration in KIB or an issue somewhere in Kubeadm and/or Kubelet.

For now though, this fix would need to be verified with every minor version bump.

faiq (approved):

PR #55, opened 2021-08-16
fix: remove refs to subscription manager

faiq:
Image builder, which this tool is based off of states that we need subscription manager to build rhel images. This PR removes that dependency by using the public key and repository directly. 

RHEL 8 GPG key bits taken from here https://github.com/ansible/ansible/issues/71634#issuecomment-687229416

### DIFF
--- a/ansible/molecule/ec2_full/molecule.yml
+++ b/ansible/molecule/ec2_full/molecule.yml
@@ -39,6 +39,20 @@ platforms:
     region: us-east-1
     instance_type: t3.small
     ssh_user: ec2-user
+  - name: konvoyimage-rhel7.9-${USER:-ci}-${HOSTNAME:-local}
+    image_search_name: "RHEL-7.9_HVM-*"
+    image_search_owner: "309956199498"
+    region: us-east-1
+    instance_type: p2.xlarge
+    spot_price: false
+    ssh_user: ec2-user
+  - name: konvoyimage-rhel8.4-${USER:-ci}-${HOSTNAME:-local}
+    image_search_name: "RHEL-8.4.0_HVM-*"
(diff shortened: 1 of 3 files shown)


ansible/roles/sysprep/tasks/redhat.yml
```diff
@@ -26,18 +26,6 @@
     masked: true
   when: ansible_memory_mb.swap.total != 0
```
tillt, line 28: I don't really understand this - why it was there in the first place.

tillt (commented):
Next step hooking it up in the molecule suite?
https://github.com/mesosphere/konvoy-image-builder/blob/2cab82639779a8c0dd7cd1e7f2e4beadb10c7fb3/ansible/molecule/ec2_full/molecule.yml

ansible/molecule/ec2_full/molecule.yml
```diff
@@ -39,6 +39,20 @@ platforms:
     region: us-east-1
     instance_type: t3.small
     ssh_user: ec2-user
+  - name: konvoyimage-rhel7.9-${USER:-ci}-${HOSTNAME:-local}
+    image_search_name: "RHEL-7.9_HVM-*"
+    image_search_owner: "309956199498"
+    region: us-east-1
+    instance_type: p2.xlarge
```
tillt, line 46: Why do we need GPU nodes for these tests? IIUC then t3.small should do, like above for centos.

tillt (approved):
Thanks @faiq!

ansible/molecule/ec2_full/molecule.yml
faiq: oops, nice catch

PR #58, opened 2021-08-18
fix: use dev/xvda as default volume for flatcar

tillt:
pulled out of the gpu feature branch to get it landed asap.

The upstream/base image is not offering an `/dev/sda1` volume, instead just a `xvda`. We need to match the upstream image block device setup.


faiq:
https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ami/flatcar.json#L14 <- used upstream as well

faiq:
https://kinvolk.io/docs/flatcar-container-linux/latest/installing/cloud/aws-ec2/#instance-storage

PR #61, opened 2021-08-20
feat: adds a rhel 8 packer build

faiq:
We previously were missing a way to build rhel 8 images for konvoy2. This creates a configuration file for packer builds to happen. Not sure what else needs to be changed

ran command like 

` bin/konvoy-image build --region us-west-2 images/ami/rhel-8.yaml`


dkoshkin (approved):
This looks fine, but I am curious why this was missing in the first place.

faiq:
@dkoshkin I think its because we copied a lot from image-builder upstream and as far as I can tell there is no packer config for rhel AMIs.

jkoelker (approved):

tillt (approved):

ansible/roles/providers/tasks/main.yml
```diff
@@ -88,6 +88,15 @@
...
   when: ansible_os_family == "Debian"
 
+- name: set cloudinit feature flags for Red Hat
+  copy:
+    src: usr/lib/python3/dist-packages/cloudinit/feature_overrides.py
+    dest: "{{ sysusr_prefix }}/lib/python3.6/site-packages/cloudinit/feature_overrides.py"
+    owner: root
+    group: root
+    mode: 0644
+  when: ansible_os_family == "RedHat"
```
jkoelker, line 98: We can just add this to the above `when` condition:

```
when: ansible_os_family in ["Debian", "RedHat"]
```

faiq: Notice the very slight difference between the paths dist vs site

ansible/roles/providers/tasks/main.yml
```diff
@@ -79,14 +78,34 @@
...
+
 - name: set cloudinit feature flags
   copy:
     src: usr/lib/python3/dist-packages/cloudinit/feature_overrides.py
     dest: "{{ sysusr_prefix }}/lib/python3/dist-packages/cloudinit/feature_overrides.py"
     owner: root
     group: root
     mode: 0644
-  when: ansible_os_family == "Debian"
+  when: ansible_os_family == "Debian" and  cloud_init_version.stdout  | float > 20.0
```
faiq, line 93: This here is a special case. Cloud init comes gets installed in dist-packages instead of site-packages.

https://wiki.debian.org/Python#Deviations_from_upstream

PR #66, opened 2021-08-23
feature: nvidia flatcar gpu

tillt:
- tested via `molecule test -s ec2_full --destroy=never` and then checked the resulting flatcar instance directly via SSH
  -  `sudo ctr image pull docker.io/nvidia/cuda:11.0-base`
  - `sudo ctr run --rm --gpus 0 -t docker.io/nvidia/cuda:11.0-base cuda-11.0-base nvidia-smi`
- still has plenty of FIXMEs and other rough edges

**ready for high level reviews or even just recommendations**


ansible/roles/gpu/files/Dockerfile.libnvidia
```diff
@@ -0,0 +1,42 @@
...
+        gnupg2 \
+        libcap-dev \
+        libelf-dev \
+        libseccomp-dev \
+        lintian \
+        lsb-release \
+        m4 \
+        pkg-config \
+        xz-utils && \
+    rm -rf /var/lib/apt/lists/*
```
asekretenko, line 28: What is the exact purpose of this `rm`? Is this necessary to prevent automatic updates during the`libnvidia-container`  build and, if yes, then is it really sufficient? I would expect that `/etc/apt/sources*` has to be removed as well (or something similar - removing the responsible cron job, etc ), otherwise `apt update` will happily recreate these.

tillt: > What is the exact purpose of this `rm`? 
Unsure myself - taken from the original

> Is this necessary to prevent automatic updates during the`libnvidia-container` build and, if yes, then is it really sufficient? I would expect that `/etc/apt/sources*` has to be removed as well (or something similar - removing the responsible cron job, etc ), otherwise `apt update` will happily recreate these.
Great points indeed.

ansible/roles/gpu/templates/config.toml.tmpl
```diff
@@ -0,0 +1,157 @@
+version = 2
```
jimmidyson, line 1: Do we really need to overwrite the whole `config.toml` or can we use separate config files just to override GPU stuff (this should be supported by containerd ootb)?

jimmidyson (approved):
Approving based on my very sketchy knowledge of what's going on here, but can't see any obvious issue.

ansible/roles/gpu/templates/usr/share/containers/oci/hooks.d/oci-nvidia-hook.json
```diff
@@ -0,0 +1,15 @@
...
+        "args": ["nvidia-container-toolkit", "prestart"],
+        "env": [
+            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+        ]
+    },
+    "when": {
+        "always": true,
+	"commands": [".*"]
+    },
+    "stages": ["prestart"]
```
tillt, line 14: This is not used in our context but has to exist IIUC.

ansible/roles/gpu/templates/usr/libexec/oci/hooks.d/oci-nvidia-hook
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" exec /usr/bin/nvidia-container-toolkit "$@"
```
tillt, line 2: Not used in our context but has to exist IIUC.

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
```diff
@@ -1,18 +1,197 @@
...
+      /opt/lib/x86_64-linux-gnu/libcrypt.so         \
+      /opt/lib/x86_64-linux-gnu/libcidn.so          \
+      /opt/lib/x86_64-linux-gnu/libBrokenLocale.so  \
+      /opt/lib/x86_64-linux-gnu/libz.so             \
+      /opt/lib/x86_64-linux-gnu/libanl.so
+
+- name: run ldconfig
+  command: ldconfig /opt/lib/x86_64-linux-gnu
+
+- name: load nvidia modules
```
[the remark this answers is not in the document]
tillt: That is indeed a prerequisite of the build - more comments incoming...

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
```diff
@@ -1,18 +1,197 @@
...
+  command: "rm /opt/lib/x86_64-linux-gnu.tar.gz"
+
+- name: extract bins into /opt/bin
+  command: "tar xzf /opt/bin/nvidia-bins.tar.gz --directory /opt/bin --strip-components 2"
+
+- name: remove nvidia-bins tarball
+  command: "rm /opt/bin/nvidia-bins.tar.gz"
+
+# FIXME(tillt): consider not packing them in the first place
+- name: kill dead libs in /opt/lib/x86_64-linux-gnu
```
dkoshkin, line 56: does "kill" have a special meaning here, is that a term used somewhere else?
```suggestion
- name: remove dead libs in /opt/lib/x86_64-linux-gnu
```

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
```diff
@@ -1,18 +1,197 @@
...
+    state: directory
+
+- name: create /opt/var/log directory
+  file:
+    path: "/opt/var/log/"
+    state: directory
+
+# build kernel modules
+- name: load loop modules
+  command: modprobe -d / -a loop i2c_core ipmi_msghandler
```
dkoshkin, line 14: Can we not use the `modprobe` module here?

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
```diff
@@ -1,18 +1,197 @@
...
+      /opt/lib/x86_64-linux-gnu/libcidn.so          \
+      /opt/lib/x86_64-linux-gnu/libBrokenLocale.so  \
+      /opt/lib/x86_64-linux-gnu/libz.so             \
+      /opt/lib/x86_64-linux-gnu/libanl.so
+
+- name: run ldconfig
+  command: ldconfig /opt/lib/x86_64-linux-gnu
+
+- name: load nvidia modules
+  command: modprobe -d /opt/nvidia/460.73.01/ -a nvidia nvidia-uvm nvidia-modeset
```
dkoshkin, line 83: Can we not use the `modprobe` module here?

ansible/vars/flatcar/flatcar.yaml
```diff
@@ -8,8 +8,6 @@ sysusr_prefix: /opt
 sysusrlocal_prefix: /opt
 systemd_prefix: /etc/systemd
 
-containerd_cri_socket: /run/docker/libcontainerd/docker-containerd.sock
-
```
dkoshkin, line 12: Why is this being removed? Is it being defined somewhere else?

dkoshkin (approved):
I think you are the closest to this @tillt and I wouldn't know if the steps are correct but the Ansible code LGTM.

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
tillt: The entire thing is very "sloppy". 

We are first getting a bunch of libraries from inside the container. Then we extract that tarball. The results will contain a bunch of broken links. Link destinations that existed within the container but not outside. Those have to get cleared to not completely get the system broken.

ansible/vars/flatcar/flatcar.yaml
tillt: Yes, the default comes from https://github.com/mesosphere/konvoy-image-builder/blob/main/ansible/group_vars/all/defaults.yaml#L22

tillt: And https://github.com/mesosphere/konvoy-image-builder/pull/66/files#diff-f1a7c1d253313ff70bf0da17787af9de98657330662cbb515b6fe38c1ded0bbcR15 then makes sure whoever still expects a docker socket, will find it.

tillt:
Thanks guys for the feedback. I will address that ASAP - for now lets not rush but wait for my update.

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
tillt: That did not work for me for reasons that are beyond me.

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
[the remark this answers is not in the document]
tillt: Just a check that they can be loaded. Comments incoming

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
tillt: Added a comment describing the intend.

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
tillt: Should try again...

ansible/roles/gpu/templates/config.toml.tmpl
tillt: No, we dont - the default one does it (is identical by now).

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
tillt: No, ansible modprobe only supports loading kernel modules from standard locations - we cant provide a directory.

ansible/roles/gpu/tasks/nvidia-gpu-Flatcar.yaml
tillt: ...actually no. I will leave a comment on top of this one indicating that we might wanna try again or find out why this one does not work. IIUC then specifying `/` means it will look at standard locations, thus ansible modprobe should be able to get it done as well. But it did not in my and Jan's testing.

PR #70, opened 2021-08-24
feat: adds instance type override option to build app

tillt:
We need to be able to make sure that flatcar images are built on p2.xlarge instances. 

I would like to avoid the command line option and instead have this somehow based on the parameters given. Don't know how though. In this specific case, we need to enable p2.xlarge when a user selected flatcar and enabled gpu overrides.


tillt:
tested running
```
dist/konvoy-image_linux_amd64/konvoy-image build --region us-west-2 images/ami/flatcar.yaml --aws-instance-type p2.xlarge  --overrides overrides/nvidia.yaml
```

Screenshot 2021-08-24 at 19 09 11

samvantran (commented):
👍

faiq (approved):

PR #82, opened 2021-09-05
fix: restart containerd upon reconfig

some-things:
Upon containerd reconfiguraiton, we should restart containerd so that it picks up the new configuration. This PR restarts containerd when the config.toml is updated on RedHat based OSes.

Flatcar will require additional changes to implement this.

### DIFF
--- a/ansible/roles/packages/tasks/install-flatcar.yaml
+++ b/ansible/roles/packages/tasks/install-flatcar.yaml
@@ -13,7 +13,8 @@
   template:
     src: "config.toml.tmpl"
     dest: "/etc/containerd/config.toml"
-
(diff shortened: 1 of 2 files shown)


jkoelker (approved):

some-things:
@jkoelker I was reviewing the way that handlers work and I think we should be fine with this change for Flatcar as well, so I've pushed a commit for that as well.

jkoelker (approved):

jkoelker:
> @jkoelker I was reviewing the way that handlers work and I think we should be fine with this change for Flatcar as well, so I've pushed a commit for that as well.

\o/

PR #126, opened 2021-11-08
chore: refactor build to use epel params

faiq:
Allows the user to specify an epel source that can be different from the upstream ones. This allows the user to download packages from their own local mirror or yum repository

### DIFF
--- a/images/common.yaml
+++ b/images/common.yaml
@@ -24,6 +24,7 @@ sysusrlocal_prefix: /usr/local
 epel_centos_7_rpm: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
 epel_centos_8_rpm: https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
 epel_centos_8_rpm_gpg_key: https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8
+epel_centos_7_rpm_gpg_key: https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-7
 
 
 kubernetes_goarch: amd64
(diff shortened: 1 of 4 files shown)


images/common.yaml
```diff
@@ -21,6 +21,10 @@ containerd_cri_socket: /run/containerd/containerd.sock
 systemd_prefix: /usr/lib/systemd/site-packages
 sysusr_prefix: /usr
 sysusrlocal_prefix: /usr/local
+epel_centos_7_rpm: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+epel_centos_8_rpm: https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+epel_centos_8_rpm_gpg_key: https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8
```
tillt, line 26: Out of curiosity; why do we not add a centos7 GPG key for EPEL? Are we not validating those?

jkoelker (approved):
\o/

images/common.yaml
jkoelker: In EPEL the signing key is part of the release rpm, I know originally the epel repo rpm wasn't signed, although it looks like it is being signed now:

```
# rpm -qpi https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm | grep Signature
warning: https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm: Header V4 RSA/SHA256 Signature, key ID 352c64e5: NOKEY
Signature   : RSA/SHA256, Sat Sep  4 17:49:16 2021, Key ID 6a2faea2352c64e5
```

Good catch we should validate this as well now too.

faiq: Great catch. I just added that in.

tillt (approved):

PR #128, opened 2021-11-13
feat: adds RHEL 8.4 packer config

tillt:


faiq:
Thank you!

Can you add a make target as well please?

PR #132, opened 2021-11-15
chore: remove double dash from fips.yaml

tillt:
Without this, the konvoy2 build-image.sh script will possibly create an override.yaml that has a `--` in its middle.

### DIFF
--- a/overrides/fips.yaml
+++ b/overrides/fips.yaml
@@ -1,4 +1,3 @@
----
 fips:
   enabled: true
   etcdImageTag: v3.4.13_fips


dkoshkin (approved):

joejulian:
Can we not fix the script so that it works with valid and proper yaml?

tillt:
@joejulian you are right, we could - this is where we had to fix it: https://github.com/mesosphere/konvoy2/pull/834

samvantran:
If this works https://github.com/mesosphere/konvoy-image-builder/pull/132#issuecomment-969140111 then let's close this PR

faiq (changes requested):
hesitant to merge this. from my understanding all of the overrides get squashed into one override file as part of the `konvoy-image` build and provision commands

joejulian:
> hesitant to merge this. from my understanding all of the overrides get squashed into one override file as part of the konvoy-image build and provision commands
The expected output is a single yaml document, not a multi-document. Instead of this change, we've changed the script that assembles the override yaml to strip out the document breaks so this change is no longer needed.

PR #136, opened 2021-11-17
ensure pause images

jr0d:


dkoshkin (approved):

faiq:
This is fine but shouldn't we take the hard coded k8s.gcr.io thing out?

faiq (approved):

dkoshkin:
> This is fine but shouldn't we take the hard coded k8s.gcr.io thing out?

Great question @faiq! There is a mystery `k8s.gcr.ip/pause:3.2` image that get pulled, leaving it here until we can figure out why its being pulled.

PR #141, opened 2021-11-22
expose ami-users and ami-groups packer variables

jr0d:
This is a blocker for gov sector. 

images now default to being private


faiq (approved):

cmd/konvoy-image/cmd/args.go
```diff
@@ -16,10 +16,52 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
...
+		&userArgs.SourceAMI,
+		"source-ami",
+		"",
+		"a specific ami available in the builder region to source from",
+	)
+	fs.StringVar(
+		&userArgs.AMIFilterName,
+		"source-ami-filter-name",
+		"",
+		"a ami name filter on for selecting the source image",
```
dlipovetsky, line 41:
```suggestion
		"an AMI name filter used to select the source image",
```

cmd/konvoy-image/cmd/args.go
```diff
@@ -16,10 +16,52 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
...
+		&userArgs.AMIRegions,
+		"ami-regions",
+		[]string{},
+		"a list of regions to publish amis",
+	)
+	fs.StringVar(
+		&userArgs.SourceAMI,
+		"source-ami",
+		"",
+		"a specific ami available in the builder region to source from",
```
dlipovetsky, line 35:
```suggestion
		"the ID of the AMI to use as the source; must be in the builder region",
```

cmd/konvoy-image/cmd/args.go
```diff
@@ -16,10 +16,52 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
...
-	fs.StringArrayVar(&userArgs.AMIRegions, "ami-regions", []string{}, "a list of regions to publish amis")
-	fs.StringVar(&userArgs.SourceAMI, "source-ami", "", "a specific ami available in the builder region to source from")
-	fs.StringVar(&userArgs.AMIFilterName, "source-ami-filter-name", "", "a ami name filter on for selecting the source image")
-	fs.StringVar(&userArgs.AMIFilterOwner, "source-ami-filter-owner", "", "only search AMIs belonging to this owner id")
-	fs.StringVar(&userArgs.AWSInstanceType, "aws-instance-type", "", "an instance type available in the builder region to work on")
+	fs.StringVar(
+		&userArgs.AWSBuilderRegion,
+		"region",
+		"",
+		"the aws region to run the builder",
```
dlipovetsky, line 23:
```suggestion
		"the region in which to build the AMI",
```

cmd/konvoy-image/cmd/args.go
```diff
@@ -16,10 +16,52 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
...
+		&userArgs.AMIFilterName,
+		"source-ami-filter-name",
+		"",
+		"a ami name filter on for selecting the source image",
+	)
+	fs.StringVar(
+		&userArgs.AMIFilterOwner,
+		"source-ami-filter-owner",
+		"",
+		"only search AMIs belonging to this owner id",
```
dlipovetsky, line 47:
```suggestion
		"a filter that restricts the source AMI to ones with this owner ID"  
```

cmd/konvoy-image/cmd/args.go
```diff
@@ -16,10 +16,52 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
...
+	)
+	fs.StringVar(
+		&userArgs.AWSInstanceType,
+		"aws-instance-type",
+		"",
+		"an instance type available in the builder region to work on",
```
dlipovetsky, line 53:
```suggestion
		"instance type used to build the AMI; the type must be present in the region in which the AMI is built",
```

cmd/konvoy-image/cmd/args.go
```diff
@@ -16,10 +16,52 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
...
+	)
+	fs.StringArrayVar(
+		&userArgs.AMIUsers,
+		"ami-users",
+		[]string{},
+		"a list AWS user accounts which are allowed use the image",
```
dlipovetsky, line 59: IIUC, to use multiple values, you can't pass a comma-separate list, but instead must invoke the flag multiple times. In that case, `--ami-user` makes more sense, e.g.,:
```
--ami-user foo --ami-user bar
```

I think the same reasoning applies to `--ami-groups`. It also applies to `--overrides`, but changing that would break compatibility, I suppose.

jr0d: @dlipovetsky Yeah, I agree. This happened originally due to my limited understanding of how cobra handled StringArray variables. I'll follow this PR and replace them with a StringSlice, which expects a "," separated string for a  value.

jr0d:
e2e already passed, only docs changes since.

cmd/konvoy-image/cmd/args.go
dlipovetsky: I'm right there with you. Very unintuitive :disappointed:

PR #145, opened 2021-11-23
ansible: remove swap from fstab during sysprep

jkoelker:
Remove swap lines from fstab during sysprep phase so provisioning physical hosts that may have it setup during the install will not fail the setup later when kubelet tries to start.

### DIFF
--- a/ansible/roles/sysprep/tasks/main.yml
+++ b/ansible/roles/sysprep/tasks/main.yml
@@ -187,3 +187,21 @@
   loop:
     - path: /root/.bash_history
     - path: "/home/{{ ansible_env.SUDO_USER }}/.bash_history"
+
+- name: Remove swap from /etc/fstab
+  mount:
+    name: "{{ item }}"
+    fstype: swap
+    state: absent
+  with_items:
+    - swap
+    - none
+
+- name: Check if swap is on
+  command: swapon -s
+  register: swapon
+  changed_when: false
+
+- name: Disable Swap
+  command: swapoff -a
+  when: swapon.stdout


msdolbey (approved):
👍 
No we're cooking with gas

dkoshkin (approved):

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -187,3 +187,21 @@
   loop:
     - path: /root/.bash_history
     - path: "/home/{{ ansible_env.SUDO_USER }}/.bash_history"
+
+- name: Remove swap from /etc/fstab
```
dkoshkin, line 191: Is this the same as https://github.com/mesosphere/konvoy-image-builder/pull/145/files#diff-fd785d2c81c482ab765a476d093e24efd4f371d54d36a9f0145f04d73db82892R172-R181?

dkoshkin: Also https://github.com/mesosphere/konvoy-image-builder/blob/125cb5b2057b61f36d9c1ed1f5df2f663ad26e41/ansible/roles/sysprep/tasks/redhat.yml#L22-L27

jkoelker: I'm not sure what that systemd service is, but it appears to be something specific to azure https://github.com/Azure/WALinuxAgent/issues/1971

jkoelker: >Is this the same as https://github.com/mesosphere/konvoy-image-builder/pull/145/files#diff-fd785d2c81c482ab765a476d093e24efd4f371d54d36a9f0145f04d73db82892R172-R181?

Those are removing common locations for swap files

jr0d (approved):

PR #149, opened 2021-11-29
use stable repo

jr0d:


ansible/group_vars/all/system.yaml
```diff
@@ -25,9 +25,9 @@ docker_rpm_gpg_key_url: "{{ docker_rpm_repository_url }}/gpg.asc"
 docker_rpm_suse_repository_url: "https://packages.d2iq.com/konvoy/stable/linux/opensuse/leap/{{ ansible_distribution_major_version|int }}/x86_64"
 docker_rpm_suse_gpg_key_url: "{{ docker_rpm_suse_repository_url }}/gpg.asc"
 
-docker_deb_repository_url: "https://konvoy-packages-test.s3.amazonaws.com/stable/linux/{{ ansible_distribution |lower }}/{{ ansible_distribution_release }}"
+# docker_deb_repository_url: "https://konvoy-packages-test.s3.amazonaws.com/stable/linux/{{ ansible_distribution |lower }}/{{ ansible_distribution_release }}"
 ## focal does not yet have a promote to staging job. Need to create one and wait for the 24 hour sync to production s3
```
samvantran, line 29: do we want to keep this comment? this is no longer true, right?

jr0d: We go back and forth.. I think a better solution might be to introduce a new variable and a flag to switch between testing/staging/stable modes.

joejulian (approved):

ansible/group_vars/all/system.yaml
jr0d: https://jira.d2iq.com/browse/D2IQ-81791

PR #155, opened 2021-12-08
fix: update for upcoming linter bump

tillt:
The new linter does not like the word "node" as used. Fixing that and updating the GPU supported platforms in the repo docs.


faiq:
Let's solve this another way. Preferably through a file described here 

https://github.com/textlint/textlint-filter-rule-allowlist

tillt:
@faiq updated.

faiq (approved):
Thank you!

PR #156, opened 2021-12-11
feat: disable firewalld and flush Iptable rules

faiq:
Allows the user to disable firewalld as a last step in system prep. The default will be to keep firewalld running. 

JIRAs: 
- https://jira.d2iq.com/browse/D2IQ-82085
- https://jira.d2iq.com/browse/D2IQ-82086
- https://jira.d2iq.com/browse/D2IQ-82087

### DIFF
--- a/ansible/roles/networking/tasks/main.yaml
+++ b/ansible/roles/networking/tasks/main.yaml
@@ -36,6 +36,12 @@
       state: present
       reload: yes
 
+  - name: Flush Iptable rules
+    iptables:
+      flush: true
+    when:
+      - flush_iptables_rules | default(true)
+
   - name: add iptable ALLOW rules for control-plane
     iptables:
       chain: INPUT
@@ -44,7 +50,7 @@
       jump: ACCEPT
       comment: "{{ item.comment }}"
     with_items:
-      - port: "{{ apiserver.secure_port }}"
+      - port: "{{ apiserver_port }}"
         comment: "Konvoy: kube-apiserver --secure-port"
       - port: 10250
         comment: "Konvoy: kubelet --port"
--- a/ansible/roles/networking/tasks/main.yaml
+++ b/ansible/roles/networking/tasks/main.yaml
@@ -36,11 +36,23 @@
       state: present
       reload: yes
 
-  - name: Flush Iptable rules
-    iptables:
-      flush: true
+  - name: Check if iptables service exists
+    command: systemctl cat iptables
+    check_mode: no
+    register: iptables_exists
+    changed_when: False
+    failed_when: iptables_exists.rc not in [0, 1]
     when:
-      - flush_iptables_rules | default(true)
+      - disable_iptables | default(true)
+
+  - name: Stop and disable defualt iptables service
+    systemd:
+      name: iptables
+      enabled: false
+      masked: false
+      state: stopped
+    when:
+      - iptables_exists.rc == 0
 
   - name: add iptable ALLOW rules for control-plane
     iptables:
--- a/ansible/roles/sysprep/tasks/main.yml
+++ b/ansible/roles/sysprep/tasks/main.yml
@@ -205,3 +205,22 @@
 - name: Disable Swap
   command: swapoff -a
   when: swapon.stdout
+
+- name: Check if firewalld service exists
+  command: systemctl cat firewalld
+  check_mode: no
+  register: firewalld_exists
+  changed_when: False
+  failed_when: firewalld_exists.rc not in [0, 1]
+  when:
+    - disable_firewalld | default(true)
+
+- name: Stop and disable firewalld service
+  systemd:
+    name: firewalld
+    enabled: false
+    masked: false
+    state: stopped
+  when:
+    - disable_firewalld | default(true)
+    - firewalld_exists.rc == 0


ansible/roles/sysprep/tasks/main.yml
```diff
@@ -188,6 +188,7 @@
     - path: /root/.bash_history
     - path: "/home/{{ ansible_env.SUDO_USER }}/.bash_history"
 
+
```
alejandroEsc, line 191: is this an accidental extra line?

alejandroEsc (approved):
lgtm, just one possible nit?

faiq:
This is the jira that describes the issues

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -205,3 +206,11 @@
...
   command: swapoff -a
   when: swapon.stdout
+
+- name: Disable firewalld if flag is set
+  systemd:
+    name: firewalld
+    enabled: false
+    masked: false
+    state: stopped
+  when: firewalld_disabled
```
dkoshkin, line 216: Konvoy 1.x [documented](https://github.com/mesosphere/konvoy/blob/4ec8d8b32c72c6c8993d5316a23c61d4d5e94169/docs/site/include/os-svc-nodes.tmpl#L4) and [disabled](https://github.com/mesosphere/konvoy/blob/8d174ec0634d2d21728ea457e511de2adc57fa6c/ansible/roles/firewalld/tasks/main.yaml) firewalld. To me KIB not doing the same is bug.

Thoughts on disabling it by default but letting users keep it enabled by setting (`disabled_firewalld:true` in an override)?
```suggestion
  when: disable_firewalld
```

faiq: @some-things suggested we do it this way and leave it to be the user's responsibility to disable it.

some-things: I'm okay with enabled or disabled by default, as long as the user has the ability to toggle it.

ansible/roles/networking/tasks/main.yaml
```diff
@@ -36,6 +36,12 @@
       state: present
       reload: yes
 
+  - name: Flush Iptable rules
+    iptables:
+      flush: true
```
dkoshkin, line 41: From https://github.com/mesosphere/konvoy/blob/8d174ec0634d2d21728ea457e511de2adc57fa6c/ansible/_flush-iptables.yaml#L9-L16

And CAPPP https://github.com/mesosphere/cluster-api-provider-preprovisioned/blob/7e4c707050a90c33b10471c954ccc211a010b237/preprovisioned/preprovisioned.go#L280-L282
```suggestion
  - name: Flush Iptable rules
    iptables:
      table: "{{ item }}"
      flush: true
    with_items:
      - nat
      - raw
```

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -205,3 +205,11 @@
...
   command: swapoff -a
   when: swapon.stdout
+
+- name: Stop and disable firewalld service
+  systemd:
+    name: firewalld
+    enabled: false
+    masked: false
+    state: stopped
+  when: disable_firewalld
```
dkoshkin, line 215: Do we need?
```suggestion
  when: disable_firewalld | default(false)
```

ansible/roles/sysprep/tasks/main.yml
dkoshkin: I suggested to disable by default because thats what Konvoy1 was doing.
We know that having it enabled will prevent Kubernetes from working unless allow certain rules are added (we can't really know fully as these will depend on user k8s workloads too).

For more sophisticated users, the `disable_firewalld` can be used to enable after these rules have been applied.

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -205,3 +205,11 @@
 - name: Disable Swap
   command: swapoff -a
   when: swapon.stdout
+
+- name: Stop and disable firewalld service
```
dkoshkin, line 209: You may need to first check if the Service exists

https://github.com/mesosphere/konvoy/blob/8d174ec0634d2d21728ea457e511de2adc57fa6c/ansible/roles/firewalld/tasks/main.yaml#L2-L9

ansible/roles/sysprep/tasks/main.yml
faiq: Does this make `disable_firewalld` default to `true`? Sorry I'm not that versed in ansible.

ansible/roles/sysprep/tasks/main.yml
faiq: This is a good idea. I'll copy some of the logic there. Good eye.

faiq: Check to see if the logic added addresses it adequately please

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -205,3 +205,22 @@
 - name: Disable Swap
   command: swapoff -a
   when: swapon.stdout
+
+- name: check if firewalld service exists
```
dkoshkin, line 209: nit:
```suggestion
- name: Check if firewalld service exists
```

ansible/roles/networking/tasks/main.yaml
```diff
@@ -36,6 +36,16 @@
...
 
+  - name: Flush Iptable rules
+    iptables:
+      table: "{{ item }}"
+      flush: true
+    with_items:
+      - nat
+      - raw
+    when:
+      - flush_iptables_rules | default(false)
```
faiq, line 47: Should this also be `default(true)` as it was in konvoy 1

dkoshkin (approved):
Is there any checks we can add to `goss`?

ansible/roles/networking/tasks/main.yaml
dkoshkin: No that was on reset, not the initial install, we can keep it `false`.

msdolbey (approved):
👍 
I like it at least!
Yeah!

dkoshkin:
Can you please rebase to pickup the Ubuntu fix.

ansible/roles/networking/tasks/main.yaml
```diff
@@ -36,6 +36,16 @@
       state: present
       reload: yes
 
+  - name: Flush Iptable rules
+    iptables:
+      table: "{{ item }}"
+      flush: true
+    with_items:
+      - nat
+      - raw
```
dkoshkin, line 45: From this convo https://mesosphere.slack.com/archives/C01MPKVBB5E/p1639503431000900?thread_ts=1639502917.498900&cid=C01MPKVBB5E we need to flush all tables, not just `nat` and `raw`.

joejulian:
This might not come from firewalld - though I'd make sure it's disabled anyway. It might also come from `iptables.service`. It might also be valuable to stop and disable that service.

For EL distros (RHEL, CentOS, SUSE), the service looks like:
```
[Unit]
Description=IPv4 firewall with iptables
AssertPathExists=/etc/sysconfig/iptables
Before=network-pre.target
Wants=network-pre.target

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/libexec/iptables/iptables.init start
ExecReload=/usr/libexec/iptables/iptables.init reload
ExecStop=/usr/libexec/iptables/iptables.init stop
Environment=BOOTUP=serial
Environment=CONSOLETYPE=serial
StandardOutput=syslog
StandardError=syslog

[Install]
WantedBy=multi-user.target
```

Stopping this service will run `/usr/libexec/iptables/iptables.init stop`, clearing the tables.

This is the centos 8 default iptables (`/etc/sysconfig/iptables`):
```
# sample configuration for iptables service
# you can edit this manually or use system-config-firewall
# please do not ask us to add additional ports/services to this default configuration
*filter
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
-A INPUT -m state --state RELATED,ESTABLISHED -j ACCEPT
-A INPUT -p icmp -j ACCEPT
-A INPUT -i lo -j ACCEPT
-A INPUT -p tcp -m state --state NEW -m tcp --dport 22 -j ACCEPT
-A INPUT -j REJECT --reject-with icmp-host-prohibited
-A FORWARD -j REJECT --reject-with icmp-host-prohibited
COMMIT
```
which restricts access to everything but ssh.

Ubuntu doesn't seem to have such a default, afaict.

dkoshkin (approved):

PR #159, opened 2021-12-13
chore: clean up images task

faiq:
Cleans up images task by using common variable `k8s_image_registry` and removing redundant pulls


dkoshkin (approved):
IIRC @jr0d left the `k8s.gcr.io/pause` images there on purpose https://github.com/mesosphere/konvoy-image-builder/commit/a6aad4ebda80efe425fddf758b5db45fc246e43f when testing air-gapped but if those images are required, we should instead fix whatever is trying to pull these images.

faiq:
Those images should still be getting pulled. They were previously getting pulled twice. one with the variable for k8s_image_registry and one hard coded `k8s.gcr.io`.  This PR removes the `k8s.gcr.io` one.

dkoshkin:
Please rebase to pickup the Ubuntu fix.

dlipovetsky:
From @jr0d at standup: When deploying a FIPS-enabled cluster, needed the k8s.gcr.io images.

dlipovetsky:
We should test with an alternate registry

joejulian:
rebased

joejulian:
rebased

dlipovetsky:
From standup: We don't need this, now that #161 merged

PR #161, opened 2021-12-14
feat: gather images dynamically

jkoelker:
Gather images dynamically from the output of `kubeadm` and `containerd`
instead of hardcoding values. Provide the ability to override any
gathered or passed image via an override map `image_replacements`.


ansible/roles/images/tasks/main.yaml
```diff
@@ -4,15 +4,30 @@
...
+    }}"
+
+- name: set aws images
+  set_fact:
+    pull_images: "{{ pull_images + aws_images }}"
   when: packer_builder_type.startswith('amazon')
+
+- name: pull images
+  shell: "crictl pull {{ item }}"
+  loop: "{{ pull_images | replace_images(image_replacements | default({})) }}"
```
dkoshkin, line 33: Can you please explain what `replace_images` is doing.

overrides/fips.yaml
```diff
@@ -1,18 +1,23 @@
 ---
-etcd_image_tag: v3.4.13_fips.0
+k8s_image_registry: mesosphere
+
 fips:
   enabled: true
-  etcdImageTag: "{{ etcd_image_tag }}"
+
+image_replacements:
+  "k8s.gcr.io/etcd:3.4.13-0": "{{ k8s_image_registry }}/etcd:v3.4.13_fips.0"
```
jr0d, line 8: Wont we also need to replace the apiserver, scheduler, controller-manager, proxy (or bake a kubeadm init config).. It needs the build metadata, `kube-apiserver:v1.21.6_fips.0` for instance.

ansible/roles/images/tasks/main.yaml
joejulian: It's replacing `  "k8s.gcr.io/etcd:3.4.13-0"` with `"{{ k8s_image_registry }}/etcd:v3.4.13_fips.0"` from the map, `image_replacments` (see `overrides/fips.yaml` below).

overrides/fips.yaml
dkoshkin: @jkoelker please see this question ☝️

jkoelker: Yea anywhere that we are changing the tag on the image, we'll need an override entry. From above though it appeared that only `etcd`'s tag was changed, and the others relied on the setting of the the image registry. But from looking at the repos in docker hub, and now seeing the below `kubernetes_build_metadata` tag that changes the other images, yea we should iterate them all out here explicitly. Otherwise some are set via `kubernetes_image_tag` and then just the `etcd` is set statically.

However I'm kinda wondering if we should look at having the `kubeadm` that we build in fips mode emit the correct tags it expects for reproducibility. That is if we "validate" that the `fips.0` build all work together, we can get rid of the image overriding all together.

I'll see if I can track down where / how we're building and publishing to the fips repo below, and if it trivial or not to make that happen. If its not, then I'll iterate them for now.

ansible/roles/images/tasks/main.yaml
jkoelker: Correct, I added a small note for the time being, but I'm going to see if we can get rid of the `image_replacements` all together.

jkoelker: I was able to get rid of this completely and move it into the playbook. Now the `build tag` from `kubeadm` drives the whole process of setting the `fips.0` (et. al)  tag.

overrides/fips.yaml
jkoelker: Went down a few rabbit holes, but I think I found a pretty nice solution:

https://github.com/mesosphere/konvoy-image-builder/pull/161/commits/8a49551524328fc4c580cc9fe94e5b3eaf9b33fd#diff-14c84fc26cc68826bdd8c9ee2ba0b29f31537fc6fec4f03028d38b024d7b9d6cR10-R25

This lets the build tag (the `GitVersion` field from `kubeadm version`) drive the process of selecting the image tags. There is a small fixup that I do to just the `etcd` image as upstream sets a build tag on them by default (e.g. `etcd:3.4.13-0`). I am stripping the `-[0-9]+` from it before appending on the build tag from `kubeadm`.

ansible/roles/config_containerd/templates/config.toml.tmpl
```diff
@@ -62,7 +62,7 @@ imports = ["/etc/containerd/conf.d/*.toml"]
     stream_server_port = "0"
     stream_idle_timeout = "4h0m0s"
     enable_selinux = false
-    sandbox_image = "{{ default_image_registry }}{{ k8s_image_registry }}/pause:{{ pause_image_version }}"
+    sandbox_image = "{{ pause_image | default('k8s.gcr.io/pause:3.6') }}"
```
jkoelker, line 65: This is just a fall back to get `containerd` to start when `download_images` is `true` and `pause_image` is not set. The `images` role will set the correct `pause_image` and then reconfigure `containerd`.

ansible/provision.yaml
```diff
@@ -35,8 +44,13 @@
     - role: config
     - role: networking
     - role: providers
+    - role: config_containerd
     - role: images
       when: download_images
+    # NOTE(jkoelker) the images task will set the correct `pause_image` to
+    #                reconfigure containerd with.
+    - role: config_containerd
+      when: download_images
```
jkoelker, line 53: This is kinda a dirty hack. I initially attempted to move the `kubeadm` discovery into the `packages` role, but its already doing *alot* and ordering everything made it more messy.

In a future, refactor we should rethink the separation of the `packages` and `config` roles, and move all the config from the packages into it (so packages can just worry about what packages and from what repos). Then a new role after `config` to "determine" the images needed by `kubeadm` to set the facts, that images will then consume.

jkoelker: Hrm... Now that I typed that out, I can just move the `kubeadm` image fetching now into a new role, and run it before the `config` role, and merge the `config_containerd` into it. It wont fix all of the config but it will be a move in the right direction. I'll do that tomorrow night.

ansible/provision.yaml
```diff
@@ -1,4 +1,13 @@
 ---
+- hosts: all
+  name: Validate Variables
+  gather_facts: false
+  tasks:
+    - name: Ensure `pause_image` is set when `download_images` is `false`
+      fail:
+        msg: Variable `pause_image` must be set if `download_images is `false`
+      when: not download_images
```
dkoshkin, line 9: Is this still needed, this var is being set in https://github.com/mesosphere/konvoy-image-builder/pull/161/files#diff-14c84fc26cc68826bdd8c9ee2ba0b29f31537fc6fec4f03028d38b024d7b9d6cR31-R37?

ansible/roles/images/tasks/main.yaml
```diff
@@ -7,21 +7,57 @@
...
-  loop: "{{ extra_images }}"
+- name: set kubernetes build tag
+  set_fact:
+    kubernetes_build_tag: "{{ kubeadm_build_tag.stdout | trim }}"
+
+# NOTE(jkoelker) If `kubeadm` has a build tag, append that to the images.
+#                Since kubernets uses a build tag on their `etcd` images (e.g
+#                `-0` from `k8s.gcr.io/etcd:3.4.13-0`), strip it prior to
+#                appending the build tag from `kubeadm`.
+- name: determine kubernetes images
```
dkoshkin, line 23: Can we write out a temp kubeadm config setting the `k8s_image_registry` and `kubernetes_build_tag` instead? This way we'll have kubeadm print us the version thats its going to use when running `init/join`.

dkoshkin: See https://github.com/mesosphere/konvoy2/blob/main/pkg/capi/generator/testdata/cluster/aws-minimal-etcd-image-ref.yaml#L92-L95

ansible/roles/config_containerd/templates/config.toml.tmpl
jkoelker: This has been removed and we will always have a pause_image set by the time `config` is run.

ansible/provision.yaml
jkoelker: It was, since that was only being set to the correct value _after_ the `images` step, and that step only runs when `download_images` was true and we needed the right value for the first `config_containerd`. Now however, it will always be set by the time `config` (since the `version`) role is running before it, and we no longer have to check it to be set for "offline" installs.

ansible/roles/images/tasks/main.yaml
jkoelker: We can set the repository that way easy peasy, but we have a chicken and egg with the `etcd` and `coredns` tags, for example the `imageTag: v3.4.13_fips.0` for etcd requires the version, but we are trying to have `kubeadm` inform us the version, how do we know that kubeadm would want (`v3.4.13`)?

We could do a two phase where we first run it without any config and just parse out the etcd and coredns versions and remove the build tag from the etcd version (upstream adds `-0` etc when they rebuild etcd), then write the config, run it again with the registry set and get the results back.

I'll prototype that out, I think that the `sed`s are ugly enough that a bit more gymnastics like that would be justified.

Sidenote it looks like for FIPS we are not building our own coredns with boringssl (I couldn't find it on dockerhub ;( ).

dkoshkin (approved):
👏

PR #162, opened 2021-12-14
feat: Import container images from cache, then pull any missing or incomplete images from remote registries

dlipovetsky:


ansible/roles/images/tasks/check-or-pull-images.yaml
```diff
@@ -0,0 +1,16 @@
...
+# If image is not present, or is present, but some layer is missing, its name is not written to stdout. Return code is 0.
+#   > ctr -n k8s.io images check -q name==docker.io/mesosphere/pause-alpine:999; echo $?
+#   0
+- name: check existing image to ensure all content is available locally
+  shell: "ctr -n k8s.io image check -q name=={{ image_name }}"
+  register: ctr_image_check
+
+- name: pull image
+  shell: "ctr -n k8s.io image pull {{ image_name }}"
+  when: ctr_image_check.stdout == "{{ image_name }}"
```
dkoshkin, line 16:
```suggestion
  when: ctr_image_check.stdout != image_name
```
OR if there are some newlines or spaces in stdout
```suggestion
  when: 'image_name not in ctr_image_check.stdout'
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -0,0 +1,5 @@
+
+- name: import images from local cache
+  shell: "ctr -n k8s.io image import {{ item }}"
```
dkoshkin, line 3: Can it be `cmd`?
```suggestion
  cmd: "ctr -n k8s.io image import {{ item }}"
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -0,0 +1,5 @@
...
+  with_fileglob:
+  - "{{ images_cache }}/*"
```
dkoshkin, line 5: Right now, the images being generated are `tar.gz`, we should add a step in here to untar (if needed).
Similar to https://github.com/mesosphere/konvoy/blob/7d8ec7e555333cb2ef9e253aceb12f95374cc7bf/ansible/roles/packages-copy-local/tasks/main.yaml#L95-L104

dlipovetsky: Yep. Going to add this.

ansible/roles/images/tasks/load.yaml
dlipovetsky: Hmm. I see we're using this form everywhere else:

```yaml
shell:
  cmd:  "..."
```

I'll change to that form.

Or are you asking about using command module, instead of the shell module?

ansible/roles/images/tasks/load.yaml
dlipovetsky: @jr0d and I made two decisions:
1. To avoid a larger than necessary cache size, we want to decompress to `ctr`'s stdin, rather than to disk.
2. We will need to use a local binary to decompress, and since `gzip` is widely available, we'll require that.

ansible/roles/images/tasks/load.yaml
dlipovetsky: (Changed from the shell to the command module, unless I needed redirection)

ansible/roles/images/tasks/check-or-pull-images.yaml
dlipovetsky: Fixed

dlipovetsky:
Depends on #166.

Introduces two dependencies on the target machine: `/bin/bash`, `ctr`, and `gzip`.

We have to use an absolute path for the shell. I wonder if we _should_  use absolute paths for the other two tools.

ansible/roles/images/tasks/load.yaml
```diff
@@ -0,0 +1,15 @@
...
+    use_regex: false
+    patterns:
+    - "*.tar.gz"
+    file_type: file
+  register: images_cache_find_result
+
+- name: import images from files in the local image cache  # noqa no-changed-when
+  shell:
+    executable: /bin/bash
+    cmd: ctr --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
```
dkoshkin, line 14: If a file is just a `tar` will running `gzip` fail?

dkoshkin: Not needed yet.

dkoshkin (approved):

dkoshkin:
@dlipovetsky please rebase this PR to fix the e2e failure.

dlipovetsky:
Rebased on main and force-pushed.

dlipovetsky:
I've added `failed_when` and `changed_when` directives to the tasks. Please squash (or let me squash) the fixup commits before merging. Thanks!

dlipovetsky:
Rebased on main for e2e test fix, and force-pushed.

ansible/roles/images/tasks/check-or-pull-images.yaml
```diff
@@ -0,0 +1,21 @@
...
+# If image is present, its name is written to stdout. Return code is 0.
+#   > ctr --namespace k8s.io images check -q name==docker.io/mesosphere/pause-alpine:3.2; echo $?
+#   docker.io/mesosphere/pause-alpine:3.2
+#   0
+#
+# If image is not present, or is present, but some layer is missing, its name is not written to stdout. Return code is 0.
+#   > ctr --namespace k8s.io images check -q name==docker.io/mesosphere/pause-alpine:999; echo $?
+#   0
+- name: check existing image to ensure all content is available locally  # noqa no-changed-when
+  command: ctr --namespace k8s.io images check name=={{ image_name }}
```
dkoshkin, line 11: Yay tests https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_KonvoyImageBuilder_E2e/3310128?buildTab=log&focusLine=18000&linesState=409

KIB configures `crictl` with the correct socket but looks `ctr` doesn't get those configs, we need something similar to:
https://github.com/mesosphere/konvoy-image-builder/blob/main/ansible/roles/config/templates/crictl.yaml

dkoshkin: An alternative is to keep using `crictls`

```
[root@ip-10-0-71-7 bin]# crictl inspecti -q k8s.gcr.io/pause:3.4.1
{
  "status": {
    "id": "sha256:0f8457a4c2ecaceac160805013dc3c61c63a1ff3dee74a473a36249a748e0253",
    "repoTags": [
      "k8s.gcr.io/pause:3.4.1"
    ],
    "repoDigests": [
      "k8s.gcr.io/pause@sha256:6c3835cab3980f11b83277305d0d736051c32b17606f5ec59f1dda67c9ba3810"
    ],
    "size": "301268",
    "uid": null,
    "username": "",
    "spec": null
  }
}
[root@ip-10-0-71-7 bin]# crictl inspecti -q k8s.gcr.io/pause:3.4.2
FATA[0000] no such image "k8s.gcr.io/pause:3.4.2" present
```

dlipovetsky: :raised_hands: Thanks tests! Will fix now.

I am surprised, because I tested the `provision` command against an EC2 machine that uses an AMI (ami-06956df2bea3220df, "konvoy-ami-centos-7-1.20.13-1639483726") that, AFAIK, is created by konvoy-image-builder.

So I expected the EC2 instance to accurately represent the state a machine is in just after the `build` command runs.

dlipovetsky: Looked at the test results; this appears to be a problem **only** on flatcar. Correct?

dlipovetsky: I don't see how we can import an image from a tarball with `crictl`.

dkoshkin: Yes only for Flatcar.

Oh good point, need to use `ctr.`

ansible/roles/images/tasks/check-or-pull-images.yaml
```diff
@@ -0,0 +1,21 @@
...
+# If image is present, its name is written to stdout. Return code is 0.
+#   > ctr --namespace k8s.io images check -q name==docker.io/mesosphere/pause-alpine:3.2; echo $?
+#   docker.io/mesosphere/pause-alpine:3.2
+#   0
+#
+# If image is not present, or is present, but some layer is missing, its name is not written to stdout. Return code is 0.
+#   > ctr --namespace k8s.io images check -q name==docker.io/mesosphere/pause-alpine:999; echo $?
+#   0
+- name: check existing image to ensure all content is available locally  # noqa no-changed-when
+  command: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images check name=={{ image_name }}
```
dkoshkin, line 11: Don't we want this (and other references in this PR)?
```suggestion
  command: ctr --address {{ flatcar_containerd_cri_socket if ansible_os_family == "Flatcar" else containerd_cri_socket }} --namespace k8s.io images check name=={{ image_name }}
```

dlipovetsky: I might be misunderstanding the provenance of this variable, but:
- It's defined in  https://github.com/mesosphere/konvoy-image-builder/blob/ae507387b4377294e9c1289e42917b32de5c0b5b/ansible/group_vars/all/defaults.yaml#L25
- And re-defined for flatcar in https://github.com/mesosphere/konvoy-image-builder/blob/125ae41a2940350c88b00698f6dfdaaa2930e3f9/ansible/vars/flatcar/flatcar.yaml#L6
- The re-definition is applied here: https://github.com/mesosphere/konvoy-image-builder/blob/e323bdcbdf3b9ed3b90fc00edfafb978e70f2e43/ansible/provision.yaml#L26-L30

If the above is right, then we should probably fix https://github.com/mesosphere/konvoy-image-builder/blob/504e73a2dad153e8d3c1ef6bb97930bcd7502fa2/ansible/roles/config/templates/crictl.yaml#L2,
https://github.com/mesosphere/konvoy-image-builder/blob/504e73a2dad153e8d3c1ef6bb97930bcd7502fa2/ansible/roles/config/templates/kubelet-drop-in.conf#L2,
and https://github.com/mesosphere/konvoy-image-builder/blob/a6aad4ebda80efe425fddf758b5db45fc246e43f/ansible/roles/packages/templates/config.toml.tmpl#L19-L23

dlipovetsky: Turns out I was wrong. After trying to figure it out with the help of https://docs.ansible.com/ansible/2.10/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable, I just gave up and put the if/else block everywhere.

This makes me wonder whether this code has any effect at all: https://github.com/mesosphere/konvoy-image-builder/blob/e323bdcbdf3b9ed3b90fc00edfafb978e70f2e43/ansible/provision.yaml#L26-L30

PR #163, opened 2021-12-14
chore: create image bundle

alejandroEsc:


ansible/_list-images.yaml
```diff
@@ -0,0 +1,4 @@
+---
+  - hosts: localhost
+    roles:
+      - role: images
```
dkoshkin, line 4: I think we want this to be its own role instead of `images`
```suggestion
    tasks:
    - name: list images
      shell: "echo {{ item }} >> ../images.out"
      loop: "{{ control_plane_images + extra_images + aws_images }}"
```

dkoshkin: This is also not in conflict with https://github.com/mesosphere/konvoy-image-builder/pull/161/files, awesome!

alejandroEsc: Tried the above out and it fails to create the images. not sure how to fix it atm but havent dug in deep enough.

dkoshkin:
@alejandroEsc PTAL at the last 2 commits and if you are agree.

dkoshkin:
Going to rebase the PR and push a couple more fixes after https://github.com/mesosphere/konvoy-image-builder/pull/169 is merged.

dkoshkin (approved):

Makefile
```diff
@@ -485,3 +490,15 @@ release-bundle: cmd/konvoy-image-wrapper/image/konvoy-image-builder.tar.gz
 	$(MAKE) GOOS=linux release-bundle-GOOS
 	$(MAKE) GOOS=windows release-bundle-GOOS
 	$(MAKE) GOOS=darwin release-bundle-GOOS
+
+.PHONY: create-image-list
+create-image-list:
+	@rm -f images.out
+	@ansible-playbook ./ansible/list-images.yaml -e="@./overrides/image-list.yaml"
```
faiq, line 497: I don't love this. It's assuming that some version of ansible-playbook in my path is the same as your path. 

We have a docker container for developing, do you think we can fix this in a future version? 

Non blocking comment though.

faiq (approved):
Approving as this works.

Makefile
faiq: I guess we can do devkit.run WHAT=create-image-list

PR #165, opened 2021-12-15
Make ssh-agent available in the devkit container

dlipovetsky:
Make it possible to use ssh keys in the devkit container. I wasn't able to run kib without this on my Linux machine. Was it user error? Feedback welcome.

```release-note

```


dkoshkin:
This PR is referencing `SSH_AUTH_SOCK`, so this to support ssh-agent?

PR #167, opened 2021-12-15
Ansible for offline installer

jr0d:
Isolated PR containing only Ansible changes for airgapped bundle support.

### DIFF
--- a/ansible/group_vars/all/defaults.yaml
+++ b/ansible/group_vars/all/defaults.yaml
@@ -45,3 +45,6 @@ kubernetes_bins:
   - kubelet
 
 sysprep: true
+offline_local_bundle_path: ""
+offline_remote_bundle_dir: "/root/.konvoy/offline"
+offline_remote_bundle_file: "offline.tar.gz"
--- a/ansible/group_vars/all/system.yaml
+++ b/ansible/group_vars/all/system.yaml
@@ -86,3 +86,10 @@ validate:
   minimum_memory_worker_mb: 2048
   # Minimum CPU cores allocated to each machine.
   minimum_cpu_cores: 2
+
+offline:
+  local_bundle_path: "{{ offline_local_bundle_path }}"
+  remote_bundle_dir: "{{ offline_remote_bundle_dir }}"
+  remote_bundle_file: "{{ offline_remote_bundle_file }}"
+  remote_bundle_path: "{{ offline_remote_bundle_dir}}/{{ offline_remote_bundle_file }}"
+  remote_bundle_os_packages_path: "{{ offline_remote_bundle_dir }}/packages"
--- a/ansible/roles/offline/tasks/upload.yaml
+++ b/ansible/roles/offline/tasks/upload.yaml
@@ -1,0 +1,11 @@
+---
+- name: create offline directory
+  file:
+    path: "{{ offline.remote_bundle_os_packages_path }}"
+    state: directory
+
+- name: upload and extract offline bundle
+  unarchive:
+    copy: true
+    src: "{{ offline.local_bundle_path }}"
+    dest: "{{ offline.remote_bundle_os_packages_path }}"
--- a/ansible/roles/packages/tasks/main.yaml
+++ b/ansible/roles/packages/tasks/main.yaml
@@ -1,3 +1,14 @@
 ---
+- name: check for offline bundle
+  stat:
+    path: "{{ offline.remote_bundle_os_packages_path }}"
+  register: offline_path
+
+- name: set offline enabled fact
+  set_fact:
+    offline_mode_enabled: "{{ offline_path.stat.exists }}"
+
 - include: repo.yaml
+  when: not offline_path.stat.exists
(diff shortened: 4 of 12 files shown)


ansible/roles/offline/tasks/upload.yaml
```diff
@@ -0,0 +1,14 @@
...
+- name: create offline directory
+  file:
+    path: "{{ offline.remote_bundle_os_packages_path }}"
+    state: directory
+
+- name: upload and extract offline bundle
+  unarchive:
+    copy: true
+    src: "{{ offline.local_bundle_path }}"
+    dest: "{{ offline.remote_bundle_os_packages_path }}"
```
dkoshkin, line 14: Nice, we can do something similar for the images.

dkoshkin (commented):
For tasks like 
```
- name: install kubectl rpm package
  yum:
    name: "{{ 'kubectl-' + package_versions.kubernetes_rpm }}"
    state: present
    update_cache: yes
  register: result
  until: result is success
  retries: 3
  delay: 3
```
And other `yum` tasks does it not try to reach out to the internet update the cache?

Could we use `enablerepo/disablerepo ` dynamically `when: offline_path.stat.isdir is defined and offline_path.stat.isdir` to and just rely on the existing tasks to install all these packages?

ansible/roles/packages/tasks/install-redhat-offline.yaml
```diff
@@ -0,0 +1,39 @@
+---
+- name: upload offline repository
+  copy:
+    src: offline.repo
+    dest: /etc/yum.repos.d/offline.repo
+
+- name: install offline packages
```
dkoshkin, line 7: We should call out the limitation about air-gapped support and `nvidia`.

dkoshkin: Do we also need `cloud-init` packages for AWS?

dkoshkin (approved):
🎉

ansible/group_vars/all/defaults.yaml
```diff
@@ -45,3 +45,6 @@ kubernetes_bins:
   - kubelet
 
 sysprep: true
+offline_local_bundle_path: ""
+offline_remote_bundle_dir: "/root/.konvoy/offline"
```
dlipovetsky, line 49: From standup: Let's think a little bit more about the naming scheme.

There might be conflict between bundles...

(Offline introduces yet another term, can we reuse an existing one?)

dlipovetsky:
I think we can merge this, but then iterate on it so that we have a single upload mechanism that works for both package and container image bundles. I'm thinking something like:
1. Generic task that copies everything in a local folder to a remote machine folder (`$HOME/.konvoy`)
2. Specialized tasks that extract the package and container bundles on the remote machine.

ansible/group_vars/all/defaults.yaml
```diff
@@ -45,3 +45,6 @@ kubernetes_bins:
...
+offline_remote_bundle_file: "offline.tar.gz"
```
dlipovetsky, line 50: Is this used anywhere? I suspect the design changed, and this is now dead code.

ansible/group_vars/all/defaults.yaml
dlipovetsky: @dkoshkin @joejulian @faiq @supershal and I talked about this, decided to use `/opt/dkp/packages` for packages (and `/opt/dkp/images` for images. Not yet sure how to change this to fit that model, though...

ansible/group_vars/all/system.yaml
```diff
@@ -86,3 +86,10 @@ validate:
   minimum_memory_worker_mb: 2048
   # Minimum CPU cores allocated to each machine.
   minimum_cpu_cores: 2
+
+offline:
+  local_bundle_path: "{{ offline_local_bundle_path }}"
+  remote_bundle_dir: "{{ offline_remote_bundle_dir }}"
+  remote_bundle_file: "{{ offline_remote_bundle_file }}"
```
jr0d, line 93: it is used here @dlipovetsky

ansible/group_vars/all/defaults.yaml
jr0d: It is used, see my next comment

ansible/roles/packages/tasks/main.yaml
```diff
@@ -1,3 +1,14 @@
...
+  register: offline_path
+
+- name: set offline enabled fact
+  set_fact:
+    offline_mode_enabled: "{{ offline_path.stat.exists }}"
+
 - include: repo.yaml
+  when: not offline_path.stat.exists
 - include: install.yaml
+  when: not offline_path.stat.exists
```
dkoshkin, line 14: This should always run since the `yum` tasks have the `offline_mode_enabled` logic

dkoshkin:
Closing in-favor of https://github.com/mesosphere/konvoy-image-builder/pull/176

PR #180, opened 2021-12-21
Revert "build(deps): bump alpine from 3.14.2 to 3.15.0"

faiq:
Reverts mesosphere/konvoy-image-builder#148 to fix this issue https://githubmemory.com/repo/atmoz/sftp/issues/296

It appears that the keys packer is creating are not compatible with this new version of open-ssh (8.8) that is shipped with alpine. This is a quick measure until we can figure out how to get packer to create different key types that are compatible with open-ssh 8.8


dkoshkin:
Please add a description why we are doing this.

PR #185, opened 2021-12-22
feat: remove outdated packages and unarchive again on changes

dkoshkin:
While testing with John and Dan, we noticed that even when an updated tar is uploaded and unarchived.

Sample output of this running.
I ran an install with one OS package bundle, the reran with a different OS package bundle.

Notice that the outdated directory was removed and unarchived again
```
TASK [offline : create offline OS packages directory] **************************
ok: [54.218.105.67]

TASK [offline : upload OS packages bundle to remote] ***************************
changed: [54.218.105.67]

TASK [offline : remove outdated os packages] ***********************************
changed: [54.218.105.67]

TASK [offline : create offline OS packages directory after cleanup] ************
changed: [54.218.105.67]

TASK [offline : unarchive OS packages bundle on remote] ************************
changed: [54.218.105.67]
```

If I rerun with no changes, nothing is modified (as expected)
```
TASK [offline : create offline OS packages directory] **************************
ok: [54.218.105.67]

TASK [offline : upload OS packages bundle to remote] ***************************
ok: [54.218.105.67]

TASK [offline : remove outdated os packages] ***********************************
skipping: [54.218.105.67]

TASK [offline : create offline OS packages directory after cleanup] ************
skipping: [54.218.105.67]

TASK [offline : unarchive OS packages bundle on remote] ************************
ok: [54.218.105.67]
```

```release-note

```

### DIFF
--- a/ansible/roles/offline/tasks/upload.yaml
+++ b/ansible/roles/offline/tasks/upload.yaml
@@ -36,6 +36,20 @@
       copy:
         src: "{{ offline.os_packages_local_bundle_file }}"
         dest: "{{ offline.os_packages_remote_bundle_path }}/offline-repo.tar.gz"
+      register: os_packages_copied
+
+    - name: remove outdated os packages
+      file:
+        path: "{{ offline.os_packages_remote_filesystem_repo_path }}"
+        state: absent
+      when: os_packages_copied.changed
+
+    # recreate the directory as it may be deleted in the previous step
+    - name: create offline OS packages directory after cleanup
+      file:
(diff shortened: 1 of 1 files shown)


faiq (approved):
Looks good to me.

faiq:
Do we want to do something similar for container images or is that not necessary?

dkoshkin:
> Do we want to do something similar for container images or is that not necessary?

Last time we spoke about this we decided not to do anything for the images. The images wouldn't have this bug, the only negative of not cleaning up images is that the install might take longer because its loading unnecessary images into Containerd.

PR #193, opened 2021-12-28
build(lint): switch from golint to revive

joejulian:
This also disables golang linting in super-linter due to a design flaw
in super-linter.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,22 @@
+---
+name: golangci-lint
+"on":
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
+  pull_request:
+permissions:
+  contents: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2.5.2
+        with:
+          version: v1.43.0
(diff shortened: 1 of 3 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,22 @@
...
+permissions:
+  contents: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2.5.2
```
dkoshkin, line 20: Were we not already running this in some github action?

joejulian: It was part of super-linter. See https://github.com/mesosphere/konvoy-image-builder/pull/193/files#diff-12809f5a94d1480f61d671f71229f03e95fe6a715813e63dd67cd83811d0181eR6-R7

joejulian: Also
https://mesosphere.slack.com/archives/C01MPKVBB5E/p1640731291375700

> Hrm... golangci-lint, in super-linter, hasn't been passing. This is because of two things, 1. super-linter only tests 1 go file at a time, and 2. golint is deprecated and archived.
> Working on a PR for these problems.
>
> The deprecation warning has been overriding the golint errors.

dkoshkin: Ahh I get it now.

dkoshkin (approved):

samvantran (approved):

tillt (approved):

.github/workflows/golangci-lint.yml
jkoelker: It was already running it in golangci-lint:

https://github.com/mesosphere/konvoy-image-builder/blob/main/.github/workflows/lint.yml#L34-L41

just that the local make didn't do the same thing ;(

PR #195, opened 2021-12-29
feat: convert centos 7 minimal iso to docker image

supershal:
- Create base centos docker image from CentOS minimal ISO file.
- Add required dependent rpms that are missing in the minimal CentOS image
- Added `e2e.centos-7-offline` to test `offline` mode

* https://jira.d2iq.com/browse/D2IQ-82941

- created teamcity job for `offline` mode https://teamcity.mesosphere.io/viewType.html?buildTypeId=ClosedSource_KonvoyImageBuilder_E2eCentOS7offline


ansible/roles/packages/tasks/install-redhat.yaml
```diff
@@ -1,4 +1,8 @@
 ---
+
+- name: clean yum cache
+  command: yum clean expire-cache
```
supershal, line 4: yum ansible module does not support cleaning cache. Please check Note section from https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html

Makefile
```diff
@@ -513,6 +516,13 @@ ci.e2e.build.%:
 
 e2e.build.centos-7: centos7 docker.clean-latest-ami
 
+# Run os-packages-artifacts outside devkit container.
+e2e.build.centos-7-offline:
+	$(MAKE) os-packages-artifacts pip-packages-artifacts
```
supershal, line 521: artifacts created by`os-packages-artifacts` and `pip-pakages-artifacts` are unavailable to devkit container. so running them on the host.
`make os-packages-artifacts` launches another docker container (rpm-getter) which mounts the `artifacts` directory from the absolute path the host. This directory path will be unavailable to devkit container if the "rpm-getter" container is created from within devkit container. 

Another alternative is to create docker volume and share between "devkit" and "rpm-getter" container.

hack/base-images/README.md
```diff
@@ -0,0 +1,29 @@
+# Create docker image from Centos ISO
```
supershal, line 1: This is a one time process to create minimal image from ISO. not sure if it belongs in this repo.

Makefile
```diff
@@ -513,6 +516,13 @@ ci.e2e.build.%:
...
+	$(MAKE) devkit.run WHAT="make save-images"
+	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline.yaml"
```
dkoshkin, line 523: Nice use of `ADDITIONAL_OVERRIDES`, glad its already useful!

hack/base-images/README.md
dkoshkin: I think its fine to keep here.
We will want to create additional images with new versions.

hack/os-packages/Makefile
```diff
@@ -23,7 +23,7 @@ download-rpms:
 	$(MAKE) rhel_major_version=$(rhel_major_version) fips=$(fips) rpm-artifacts-getter.check
 	$(MAKE) rhel_major_version=$(rhel_major_version) fips=$(fips) packages="audit ca-certificates conntrack-tools chrony curl ebtables open-vm-tools python3-pip python-netifaces python-requests socat sysstat yum-utils yum-plugin-versionlock nfs-utils NetworkManager" download-rpm
 	# packages that are not in CentOS-7 Minimal
-	$(MAKE) rhel_major_version=$(rhel_major_version) fips=$(fips) packages="tzdata nspr nss python-chardet yum-utils nss-softokn-freebl nss-util nss-softokn nss nss-tools libxml2-python python-kitchen yum-plugin-ovl" download-rpm
+	$(MAKE) rhel_major_version=$(rhel_major_version) fips=$(fips) packages="tzdata nspr nss python-chardet yum-utils nss-softokn-freebl nss-util nss-softokn nss nss-tools  NetworkManager-libnm libdrm libpciaccess libxml2-python python-kitchen yum-plugin-ovl" download-rpm
```
dkoshkin, line 26: What happens if you remove this whole line do these packages get downloaded because they are dependencies of other packages that are being downloaded? We do not have a clear way of generating this list of images, it would be ideal if this Makefile only needed to specify the packages that are explicitly installed in the Ansible code.

dkoshkin (approved):

hack/os-packages/Makefile
supershal: I tested it in offline mode so the missing dependency is not downloaded and I had to add those. We can do `yum deplist <required rpms>` and compile list of rpm required by comparing it with rpms already available in minimum iso + `core` packages. I can give it a try.

hack/base-images/README.md
supershal: sounds good. I will add automating this script in teamcity in next couple of sprint then.

hack/base-images/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+This page documents steps required to create minmal docker image from centos 7 ISO
+
+The build scripts are slightly modified version of <https://github.com/CentOS/sig-cloud-instance-build/tree/master/docker>
+Build scripts on the `sig-cloud-instance-build` repository are used to create official centos docker images.
+
+## Prerequisites
+All steps needed to be executed in a Centos VM.
```
faiq, line 9: Do we need to copy the ks file to the new build machine as well?

hack/base-images/containerbuild.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
faiq, line 1: In a comment can you add where you borrowed this script from?

faiq (approved):
looks good

hack/base-images/README.md
supershal: yes. I have added instructions in the readme file to do the same.

hack/base-images/containerbuild.sh
supershal: Added the link to the script.

PR #205, opened 2022-01-11
chore: create RHEL 7.9 fips image for airgapped installations

supershal:
- build rhel 7.9 AMI for FIPS and airgapped mode
- refactor `os-package` building code to support multiple OS distros/versions/FIPS combinations

- https://jira.d2iq.com/browse/D2IQ-83344

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,6 +239,14 @@ rhel79: ## Build RHEL 7.9 image
 	-v ${VERBOSITY} \
 	$(if $(ADDITIONAL_OVERRIDES),--overrides=${ADDITIONAL_OVERRIDES})
 
+.PHONY: rhel79-fips
+rhel79-fips: build
+rhel79-fips: ## Build RHEL 7.9 image
+	./bin/konvoy-image build images/ami/rhel-79.yaml \
+	-v ${VERBOSITY} \
+	--overrides=overrides/fips.yaml \
+	$(if $(ADDITIONAL_OVERRIDES),--overrides=${ADDITIONAL_OVERRIDES})
+
 .PHONY: rhel79-nvidia
 rhel79-nvidia: build
 rhel79-nvidia: ## Build RHEL 7.9 image with GPU support
@@ -511,7 +519,6 @@ endif
 # Output is interleaved when run in parallel. Use --output-sync=recurse to serialize output.
 ci.e2e.build.all: ci.e2e.build.centos-7
 ci.e2e.build.all: e2e.build.centos-7-offline
-ci.e2e.build.all: e2e.build.centos-7-offline-fips
 ci.e2e.build.all: ci.e2e.build.centos-8
 ci.e2e.build.all: ci.e2e.build.ubuntu-18
 ci.e2e.build.all: ci.e2e.build.ubuntu-20
@@ -519,6 +526,7 @@ ci.e2e.build.all: ci.e2e.build.sles-15
 ci.e2e.build.all: ci.e2e.build.oracle-7
 ci.e2e.build.all: ci.e2e.build.oracle-8
 ci.e2e.build.all: ci.e2e.build.flatcar
+ci.e2e.build.all: e2e.build.rhel-7.9-offline-fips
 ci.e2e.build.all: ci.e2e.build.rhel-8-fips
 ci.e2e.build.all: ci.e2e.build.centos-7-nvidia
 ci.e2e.build.all: ci.e2e.build.centos-8-nvidia
@@ -537,10 +545,10 @@ e2e.build.centos-7-offline:
 	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline.yaml"
 	$(MAKE) docker.clean-latest-ami
 
-e2e.build.centos-7-offline-fips:
+e2e.build.rhel-7.9-offline-fips:
 	$(MAKE) os-packages-artifacts pip-packages-artifacts
 	$(MAKE) devkit.run WHAT="make save-images"
-	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline-fips.yaml"
+	$(MAKE) devkit.run WHAT="make rhel79-fips ADDITIONAL_OVERRIDES=overrides/offline-fips.yaml"
 	$(MAKE) docker.clean-latest-ami
 
 e2e.build.centos-8: centos8 docker.clean-latest-ami
--- a/Makefile
+++ b/Makefile
@@ -329,6 +329,7 @@ ci: ## CI build
 ci: dev diff
 
 .PHONY: clean
+clean: clean-artifacts
 clean: ## remove files created during build
 	$(call print-target)
 	rm -rf bin
@@ -336,7 +337,6 @@ clean: ## remove files created during build
 	rm -rf "$(REPO_ROOT_DIR)/cmd/konvoy-image-wrapper/image/konvoy-image-builder.tar.gz"
 	rm -f flatcar-version.yaml
 	rm -f $(COVERAGE)*
-	rm -rf artifacts
 	docker image rm $(DOCKER_DEVKIT_IMG) || echo "image already removed"
 
 .PHONY: generate
--- a/Makefile
+++ b/Makefile
@@ -511,6 +511,7 @@ endif
 # Output is interleaved when run in parallel. Use --output-sync=recurse to serialize output.
 ci.e2e.build.all: ci.e2e.build.centos-7
 ci.e2e.build.all: e2e.build.centos-7-offline
+ci.e2e.build.all: e2e.build.centos-7-offline-fips
 ci.e2e.build.all: ci.e2e.build.centos-8
 ci.e2e.build.all: ci.e2e.build.ubuntu-18
 ci.e2e.build.all: ci.e2e.build.ubuntu-20
@@ -534,6 +535,12 @@ e2e.build.centos-7-offline:
 	$(MAKE) os-packages-artifacts pip-packages-artifacts
 	$(MAKE) devkit.run WHAT="make save-images"
 	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline.yaml"
+	$(MAKE) docker.clean-latest-ami
+
+e2e.build.centos-7-offline-fips:
+	$(MAKE) os-packages-artifacts pip-packages-artifacts
+	$(MAKE) devkit.run WHAT="make save-images"
+	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline-fips.yaml"
 	$(MAKE) docker.clean-latest-ami
 
 e2e.build.centos-8: centos8 docker.clean-latest-ami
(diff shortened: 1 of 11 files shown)


hack/os-packages/Makefile
```diff
@@ -1,17 +1,20 @@
...
 artifacts/os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_x86_64_rpms.tar.gz:
 	cd artifacts/os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_el7_x86_64_rpms && tar -czf ../os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_x86_64_rpms.tar.gz *
 
+artifacts/os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_x86_64_fips_rpms.tar.gz:
+	cd artifacts/os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_el7_x86_64_fips_rpms && tar -czf ../os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_x86_64_fips_rpms.tar.gz *
+
 .PHONY: download-all-rpms
 download-all-rpms:
-	$(MAKE) rhel_major_version=7 fips=0 bastion=0  download-rpms
+	$(MAKE) rhel_major_version=7 fips=0 download-rpms
```
supershal, line 16: removed `bastion=0` parameter as it was not used. It must have added for future purpose. Please let me know if I should keep it and how are we going to use it in future.

hack/os-packages/Makefile
```diff
@@ -31,34 +34,40 @@ download-rpms:
 
 ifeq ($(fips), 1)
 artifacts_dir = $(shell pwd)/artifacts/os-packages_$(DEFAULT_KUBERNETES_VERSION_SEMVER)_el$(rhel_major_version)_x86_64_fips_rpms
-rpm_getter_image = $(RPM_PACKAGES_GETTER_IMAGE)-fips-$(rhel_major_version)
-getter_target = rpm-fips-artifacts-getter
+rpm_getter_image = $(RPM_PACKAGES_GETTER_IMAGE)-fips-centos:$(rhel_major_version)
```
supershal, line 37: The original code was using k8s version as tag for the rpm-getter image. IMHO since multiple k8s versions can be built against same base OS image, I changed the tag to be `major_version` of the OS.

hack/os-packages/Makefile
```diff
@@ -1,17 +1,20 @@
 RPM_PACKAGES_GETTER_IMAGE = mesosphere/rpm-artifacts-getter
-RPM_FIPS_PACKAGES_GETTER_IMAGE = mesosphere/rpm-artifacts-getter-fips
-
+CENTOS_BASE_IMAGE=mesosphere/centos:7.9.2009.minimal
```
supershal, line 2: hardcoding to centos:7.9.2009.minimal for now. this will be parameter in future iterations.

hack/os-packages/rpm-artifacts-getter/centos/7/Dockerfile
```diff
@@ -0,0 +1,8 @@
+ARG BASE_IMAGE=mesosphere/centos:7.9.2009.minimal
```
supershal, line 1: opinionated directory structure introduced: `rpm-artifacts-getter/<OS>/<MAJOR_VERSION>/` . please let me know if any other structure makes more sense.

hack/os-packages/rpm-fips-artifacts-getter/centos/7/containerd.fips.repo
```diff
@@ -0,0 +1,5 @@
+[containerd-fips]
+baseurl = https://containerd-fips.s3.us-east-2.amazonaws.com/7/x86_64
+gpgcheck = 0
+gpgkey = ""
```
supershal, line 4: Help needed: Unable to find out gpgkey URL for this repo. 
Should we change the baseURL to point to `packages.d2iq.com` for the FIPS packages for both kuberentes and containerd rpms ?

Makefile
```diff
@@ -536,6 +537,12 @@ e2e.build.centos-7-offline:
 	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline.yaml"
 	$(MAKE) docker.clean-latest-ami
 
+e2e.build.centos-7-offline-fips:
+	$(MAKE) os-packages-artifacts pip-packages-artifacts
+	$(MAKE) devkit.run WHAT="make save-images"
+	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline-fips.yaml"
```
dkoshkin, line 543: The JIRA ticket is for RHEL 7.9 FIPS. Im not sure if FIPS on a non-RHEL OS can actually valid.

What do you think on something similar to adding something similar to https://github.com/mesosphere/konvoy-image-builder/blob/b05836f3883612b4f3a8c0d01ed2a524f39f0bd4/Makefile#L212-L218 but for RHEL 7.9

```suggestion
	$(MAKE) devkit.run WHAT="make rhel7.9-fips ADDITIONAL_OVERRIDES=overrides/offline.yaml"
```

This way we don't need to duplicate any of the FIPS vars.

hack/os-packages/Makefile
dkoshkin: This change makes sense!

hack/os-packages/rpm-fips-artifacts-getter/centos/7/containerd.fips.repo
dkoshkin, line 4:
```suggestion
gpgkey = "https://containerd-fips.s3.us-east-2.amazonaws.com/7/x86_64/gpg.asc"
```

dkoshkin: https://github.com/mesosphere/konvoy/blob/76c918e77edd035ccb01e1b5ea5da8a607891841/ansible/group_vars/all/system.yaml#L112

hack/os-packages/rpm-artifacts-getter/centos/7/containerd.repo
```diff
@@ -0,0 +1,5 @@
+[konvoy-packages]
+baseurl = https://packages.d2iq.com/konvoy/stable/linux/centos/7/x86_64
+gpgcheck = 0
+gpgkey = ""
```
dkoshkin, line 4:
```suggestion
gpgkey = "https://packages.d2iq.com/konvoy/stable/linux/centos/7/x86_64/gpg.asc"
```

hack/os-packages/rpm-artifacts-getter/centos/7/kubernetes.repo
```diff
@@ -0,0 +1,5 @@
+[kubernetes]
+name=Konvoy Kubernetes package repository
+baseurl=https://kubernetes-fips.s3.us-east-2.amazonaws.com/7/x86_64
```
dkoshkin, line 3:
```suggestion
baseurl=https://packages.d2iq.com/konvoy/rpm/stable/centos/7/x86_64
```

dkoshkin (changes requested):

hack/os-packages/Makefile
dlipovetsky: Thanks for cleaning it up

Makefile
supershal: Agree. I will make the changes.

hack/os-packages/rpm-fips-artifacts-getter/centos/7/containerd.fips.repo
supershal: Thank you

hack/os-packages/rpm-artifacts-getter/centos/7/kubernetes.repo
supershal: Good catch. 👍

hack/os-packages/rpm-fips-artifacts-getter/centos/7/containerd.fips.repo
supershal: thanks for providing link to gpgkey

dkoshkin:
In your testing have you seen any issues with https://github.com/mesosphere/konvoy-image-builder/blob/b8ecfc57bf0444471e8882ebf27e8e6df6f981bb/ansible/roles/packages/tasks/install-redhat.yaml#L66-L76?

This task would not work without internet but we don't have a `when` condition skipping it.

supershal:
@dkoshkin Thanks for explaining the pre-provision use-case during standup.
This works during my testing because the image building process does not run into airgapped environment.
https://teamcity.mesosphere.io/viewLog.html?buildId=3324821&buildTypeId=ClosedSource_KonvoyImageBuilder_E2eCentOS7offlineFips&tab=buildLog&_focus=2713#_state=469
```
[01:32:00][Step 2/3]     rhel-7.9-fips: TASK [packages : install container-selinux GPG key] ****************************
[01:32:01][Step 2/3]     rhel-7.9-fips: changed: [default]
```
I can add when condition to not run this task when offline mode is enabled for this PR.

supershal:
I am thinking to create separate PR to run entire image building process in air-gapped environment. 
However i am thinking to pre-create VPC with NAT gateways that blocks traffic to internet instead of using security group.
Use this pre-created known VPC everywhere when we want to test airgapped flows on KIB and Konvoy. 
This way if any component running inside VPC does not have to change their own SG individually. Wdyt?

supershal:
Also, If i am making changes to run this in airgapped mode, I prefer to change the way we decide "airgapped enabled" mode using ansible facts and based on if a path to the offline bundle exists.
I like to change it to make it more explicit. Also change the term `offline` to `airgapped` to avoid any confusion. 
```yaml
airgapped:
  enabled: true
  os_packages_local_bundle_file: "/path/to/os/packages"
  pip_packages_local_bundle_file: "/path/to/pip/package"
  images_local_bundle_dir: /path/to/docker/images"
```
wdyt?

hack/os-packages/rpm-artifacts-getter/centos/7/kubernetes.repo
supershal: fixed

dkoshkin:
> Also, If i am making changes to run this in airgapped mode, I prefer to change the way we decide "airgapped enabled" mode using ansible facts and based on if a path to the offline bundle exists. I like to change it to make it more explicit. Also change the term `offline` to `airgapped` to avoid any confusion.
> 
> ```yaml
> airgapped:
>   enabled: true
>   os_packages_local_bundle_file: "/path/to/os/packages"
>   pip_packages_local_bundle_file: "/path/to/pip/package"
>   images_local_bundle_dir: /path/to/docker/images"
> ```
> 
> wdyt?

We want to keep 2 parts of the "airgapped" story separate. The copying of bundles to the remote host is controlled with ansible vars `airgapped.os_packages_local_bundle_file` `airgapped.pip_packages_local_bundle_file` `airgapped.images_local_bundle_dir` and then the tasks that use those bundles only check for presence of the files and not some args from the user.

This allows to be more flexible in our preporivisioned provider where a user can have their own process to place these bundles on the host and not have to set additional vars when creating a cluster.

dkoshkin (approved):

dkoshkin:
>This way if any component running inside VPC, it does not have to change their own SG individually to block internet
traffic. Do we already have such VPC?
OR do we already have a scripts to create airgapped env in AWS that I can leverage and feed it to packer at runtime?

No we do not have such a VPC and we would need to go through SRE to have it not be cleaned up by cloud-cleaner.
This VPC would probably be pretty specific to KIB, in Konvoy reusing a VPC could cause conflicts.
https://github.com/mesosphere/konvoy2/blob/main/test/e2e/suites/infra/aws-air-gapped/main.tf is the TF that konvoy uses to create an airgapped VPC.

PR #211, opened 2022-01-23
 fix(ansible): reuse roles for image saving

jkoelker:

Two commits for `os-packages` to set target defaults, as well as use a for loop for cleaning the image getting around xargs always running at least once.

Refactors `save-images` to reuse the  ansible roles and a temporary container to fetch image versions.

This is basically #210 again, but i accidentally push deleted that branch, so it close that PR :doh:

```release-note

```


ansible/roles/sysprep/tasks/main.yml
```diff
@@ -224,3 +224,15 @@
...
+
+- name: Ensure containterd is started
+  systemd:
+    name: containerd
+    enabled: true
+    state: started
+
+- name: Ensure kubelet is started
+  systemd:
+    name: containerd
```
supershal, line 236:
```suggestion
    name: kubelet
```

ansible/list-images.yaml
```diff
@@ -1,2 +1,48 @@
...
+    - add_host:
+        name: "{{ container_name }}"
+        ansible_connection: docker
+        changed_when: false
+        ansible_python_interpreter: /usr/libexec/platform-python
+
+- hosts: "{{ hostvars['localhost'].container_name }}"
+  roles:
+    - role: packages
+    - role: version
```
supershal, line 27: I think `images` role should be added here. 
Also I am not sure why `packages` role is required in this playbook.

supershal (changes requested):

ansible/roles/sysprep/tasks/main.yml
jkoelker: Whoops, bad copy pasta. Good catch. Fixed and repushed. But since we're talking about it, should we just remove the starting of `kubelet` in KIB. It will fail to start since there is no config, and the `goss` checks just check that it is present and either started or stopped.

ansible/list-images.yaml
supershal: it makes sense now. Thanks for explaining it. 
w.r.t refactoring, I think there might be more cleanup needed. We can pair up offline to refactor the roles and do it as separate PR. wdyt?

supershal (approved):

ansible/list-images.yaml
jkoelker: Yep sounds good.

PR #214, opened 2022-01-24
feat: adds an upload artifacts command to konvoy image builder

faiq:
Creates an `upload artifacts` command to upload artifacts to hosts specified in an inventory file.

https://jira.d2iq.com/browse/D2IQ-83933

I ran the command like so:

```
$ bin/konvoy-image-wrapper upload artifacts --inventory-file=ansible-inventory.yaml --container-images-dir=./artifacts/image-bundles/ --os-packages-bundle=./artifacts/os-packages_1.21.6_x86_64_rpms.tar.gz --pip-packages-bundle=./artifacts/pip-packages.tar.gz 
```

I used previous existing terraform scripts to bring up centos7 hosts. 

```release-note
feat: adds a upload artifacts command for hosts that will be used in offline mode.
```

### DIFF
--- a/cmd/konvoy-image/cmd/artifacts.go
+++ b/cmd/konvoy-image/cmd/artifacts.go
@@ -1,0 +1,33 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/mesosphere/konvoy-image-builder/pkg/app"
+)
+
+var artifactsFlags app.ArtifactsCmdFlags
+
+var artifactsCmd = &cobra.Command{
+	Use:   "artifacts",
+	Short: "upload artifacts to hosts defined in inventory-file",
+	Args:  cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		err := app.UploadArtifacts(artifactsFlags)
+		if err != nil {
+			return fmt.Errorf("failed to upload artifacts %w", err)
+		}
+		return nil
+	},
+}
+
+func init() {
+	fs := artifactsCmd.Flags()
+	fs.StringVar(&artifactsFlags.Inventory, "inventory-file", "inventory.yaml", "an ansible inventory defining your infrastructure")
+	fs.StringVar(&artifactsFlags.OSPackagesBundleFile, "os-packages-bundle", "", "path to os-packages tar file for install on remote hosts.")
+	fs.StringVar(&artifactsFlags.PIPPackagesBundleFile, "pip-packages-bundle", "", "path to pip-packages tar file"+
+		"for install on remote hosts.")
+	fs.StringVar(&artifactsFlags.ContainerImagesBundleDir, "container-images-dir", "", "path to container images for install on remote hosts.")
+}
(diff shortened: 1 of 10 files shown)


supershal:
Please add related target in `Makefile`. I think the `e2e.*.offline` targets will change to support this.

cmd/konvoy-image/cmd/artifacts.go
```diff
@@ -0,0 +1,33 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to upload artifacts %w", err)
+		}
+		return nil
+	},
+}
+
+func init() {
+	fs := artifactsCmd.Flags()
+	fs.StringVar(&artifactsFlags.Inventory, "inventory-file", "inventory.yaml", "an ansible inventory defining your infrastructure")
```
supershal, line 28: we should add shorthands notations to all commands. wdyt?

cmd/konvoy-image/cmd/artifacts.go
```diff
@@ -0,0 +1,33 @@
...
+	fs.StringVar(&artifactsFlags.OSPackagesBundleFile, "os-packages-bundle", "", "path to os-packages tar file for install on remote hosts.")
+	fs.StringVar(&artifactsFlags.PIPPackagesBundleFile, "pip-packages-bundle", "", "path to pip-packages tar file"+
```
supershal, line 30: The flags should be marked as `Required` so that it will show up as required in the `help`.

supershal:
for next iteration: Should we think about combining all artifacts into a single versioned tar if we need to provide all three type of artifacts (rpms, pip packages, images) all the time? This could prevent version mismatching between `rpms` and `images` artifacts. also it would be easier for users to manage/download one single tar file. Then this command could become simple where it could upload/sync sort of `anything` to each host in the inventory.

supershal (changes requested):

cmd/konvoy-image/cmd/artifacts.go
faiq: Good idea!

ansible/provision.yaml
```diff
@@ -29,7 +29,6 @@
     - include_vars: vars/flatcar/flatcar.yaml
       when: ansible_os_family == "Flatcar"
   roles:
-    - role: offline
```
faiq, line 32: We should actually keep this here. In the case where the user is trying to build an AMI with the offline bundles present

cmd/konvoy-image/cmd/artifacts.go
faiq: These should probably be addressed in a refactor for all of the subcommands.

supershal: sounds good.

supershal: if we are following kubectl conventions, it would be good to follow those when we refactoring flags. 
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-cli/kubectl-conventions.md#flag-conventions

supershal (approved):
this is going to be a very useful feature.

docs/cli/konvoy-image_upload_artifacts.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+### Options
+
+```
+      --container-images-dir string   path to container images for install on remote hosts.
+  -h, --help                          help for artifacts
+      --inventory-file string         an ansible inventory defining your infrastructure (default "inventory.yaml")
+      --os-packages-bundle string     path to os-packages tar file for install on remote hosts.
+      --pip-packages-bundle string    path to pip-packages tar filefor install on remote hosts.
```
JohnB64-JBV, line 16:
```suggestion
      --pip-packages-bundle string    path to pip-packages tar file for install on remote hosts.
```

JohnB64-JBV (approved):

PR #217, opened 2022-01-26
test: creates airgapped infra for tests

faiq:
Creates an airgapped infrastructure for us to run our offline packer builds in 

* https://jira.d2iq.com/browse/D2IQ-83638

```release-note
feat: allow users to specify security group id in images.yaml 
```


test/infra/aws/export.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+export AWS_VPC_ID="$("${TERRAFORM_BIN}" -chdir=.local/infra/aws output -raw vpc_id)"
+export AWS_SECURITY_GROUP_ID="$("${TERRAFORM_BIN}" -chdir=.local/infra/aws output -json | "${GOJQ_BIN}" .security_group_id.value[0])"
+export AWS_SUBNET_ID="$("${TERRAFORM_BIN}" -chdir=.local/infra/aws output -json | "${GOJQ_BIN}" .public_subnets.value[0])"
```
dkoshkin, line 8: If we are only interested in the first value can we instead just output it from terraform

Something like this:
```
output "public_subnets" {
  value = aws_subnet.konvoy_public.[0].id
}

output  "security_group_id" {
  value = aws_security_group.konvoy_ssh.[0].id
}
```

dkoshkin: This would allow us to remove `gojq`.

Makefile
```diff
@@ -16,7 +16,31 @@ GROUP_NAME ?= $(shell id -g -n)
 
 COVERAGE ?= $(REPO_ROOT_DIR)/coverage
 
-VERBOSITY ?= 6
+VERBOSITY ?= 0
```
faiq, line 19: Let me know if you'd like me to change this back. Running it locally is frustrating with it at 6

test/infra/aws/main.tf
```diff
@@ -0,0 +1,89 @@
...
+  default     = "10.0.0.0/16"
+}
+
+variable "tags" {
+  description = "Map of tags to add to all resources"
+  type        = map
+  default     = {}
+}
+
+resource "aws_vpc" "konvoy_vpc" {
```
dkoshkin, line 26: Maybe I misunderstood but I thought we weren't going to create infra for every test run?
Instead create a SecurityGroup(if its already there it would be a noop) in the default VPC and Subnet with an S3 as the TF storage.

faiq: No, you're absolutely right that is what we talked about. 

I didn't want to leave the security-group id or subnet-id in a public repo so i had to go with the templating appraoch. I'm not sure how to pull values out using an s3 backend so i did the local directory approach.

I'll try to change it more once I confirm the general idea works.

dkoshkin: Oh yeah good point about the IDs.

Instead of having this run by CI every time we could set it the IDs in CI and have another process run this when actually needed, its unlikely that the SG would be deleted by something or someone so even having docs how to run this locally to recreate the SG in the default VPC and what TC jobs need to be updated with the ID would be enough.

dkoshkin: Actually looking at the job logs it takes about a minute to create these resources, maybe we should just create it for each run.
Only thing is we may want to add a TC step at the end that runs `$(MAKE) devkit.run WHAT="make infra.aws.destroy"` to not leak resources.

Makefile
```diff
@@ -542,8 +543,13 @@ e2e.build.centos-7: centos7 docker.clean-latest-ami
 e2e.build.centos-7-offline:
 	$(MAKE) os-packages-artifacts pip-packages-artifacts
 	$(MAKE) devkit.run WHAT="make save-images"
-	$(MAKE) devkit.run WHAT="make centos7 ADDITIONAL_OVERRIDES=overrides/offline.yaml"
+	$(MAKE) devkit.run WHAT="make infra.aws.init"
+	$(MAKE) devkit.run WHAT="make infra.aws.create"
+	$(MAKE) devkit.run WHAT="make packer-custom-vpc-override.yaml"
+	$(MAKE) devkit.run WHAT="make build"
+	$(MAKE) devkit.run WHAT="./bin/konvoy-image build images/ami/centos-7.yaml -v ${VERBOSITY} --overrides=overrides/offline.yaml --overrides=packer-custom-vpc-override.yaml"
```
supershal, line 550: I faced same issue of difficulty in providing multiple override files dynamically to makefile targets. This is one way to address it https://github.com/mesosphere/konvoy-image-builder/pull/220 . PTAL

supershal (changes requested):

test/infra/aws/main.tf
faiq: Added to the centos7 one and will add as more tests begin to use this environment

Makefile
```diff
@@ -580,7 +582,7 @@ ci.e2e.build.%:
 e2e.build.centos-7: centos7 docker.clean-latest-ami
 
 # Run os-packages-artifacts outside devkit container.
-e2e.build.centos-7-offline: centos7-offline docker.clean-latest-ami
+e2e.build.centos-7-offline: centos7-offline docker.clean-latest-ami infra.aws.destroy
 
 e2e.build.rhel-7.9-offline-fips: rhel79-fips-offline docker.clean-latest-ami
```
dkoshkin, line 587: We also want to make a similar change for the `rhel-x.x-offline-fips` jobs, is that going to be done as part of this?

Makefile
```diff
@@ -580,7 +582,7 @@ ci.e2e.build.%:
 e2e.build.centos-7: centos7 docker.clean-latest-ami
 
 # Run os-packages-artifacts outside devkit container.
-e2e.build.centos-7-offline: centos7-offline docker.clean-latest-ami
+e2e.build.centos-7-offline: centos7-offline docker.clean-latest-ami infra.aws.destroy
```
dkoshkin, line 585:
```suggestion
e2e.build.centos-7-offline: centos7-offline
```

What do you think about adding new steps in TC to do this cleanup `docker.clean-latest-ami infra.aws.destroy` to not leak AMIs and infra on failed tests"

pkg/packer/manifests/aws/packer.json.tmpl
```diff
@@ -181,6 +188,7 @@
       "format_options": "{{user `goss_format_options`}}",
       "goss_file": "{{user `goss_entry_file`}}",
       "inspect": "{{user `goss_inspect_mode`}}",
+      "skip_install": false,
```
supershal, line 191: should we need to default it with `true` or read it from parameter?

supershal (changes requested):

pkg/packer/manifests/aws/packer.json.tmpl
faiq: i meant to write true :doh:

Makefile
faiq: the destroy infra step is already doing in TC. I think its worth keeping this in both in case you want to run locally

supershal (approved):

Makefile
faiq: Added! Just wanted to get this working with one OS first before delving in to more :)

ansible/roles/version/tasks/main.yaml
```diff
@@ -16,8 +16,9 @@
 #                the build tag.
 - name: determine kubernetes images
   shell: |
-    kubeadm config images list --kubernetes-version='{{ kubernetes_semver }}' --image-repository '{{ k8s_image_registry }}' |
-    sed 's/^\(.\+etcd:\)\(.\+\)-\([0-9]\+\)$/\1{{ "\2-\3" if not kubernetes_build_tag else "v\2"}}/' |
+    kubeadm config images list \
+      --kubernetes-version='{{ kubernetes_semver }}' --image-repository '{{ k8s_image_registry }}' |
+    sed 's/^\(.\+etcd:.\+\)-\([0-9]\+\)$/\1{{ "-\2" if not kubernetes_build_tag }}/' |
```
supershal, line 21: @faiq Looks like this line is rebased correctly from main. It has a fix for using `v` in version.
https://github.com/mesosphere/konvoy-image-builder/blob/main/ansible/roles/version/tasks/main.yaml#L17
```suggestion
    kubeadm config images list --kubernetes-version='{{ kubernetes_semver }}' --image-repository '{{ k8s_image_registry }}' |
    sed 's/^\(.\+etcd:\)\(.\+\)-\([0-9]\+\)$/\1{{ "\2-\3" if not kubernetes_build_tag else "v\2"}}/' |
```

supershal (changes requested):

supershal (approved):
Great work to get it work end to end in airgapped env.

ansible/roles/version/tasks/main.yaml
```diff
@@ -16,6 +16,8 @@
 #                the build tag.
 - name: determine kubernetes images
   shell: |
+    kubeadm config images list \
+      --kubernetes-version='{{ kubernetes_semver }}' --image-repository '{{ k8s_image_registry }}' |
```
dkoshkin, line 20: Maybe a bad rebase?
```suggestion
```

dkoshkin (approved):

PR #220, opened 2022-01-26
build: improvements to provide additional override files to makefile targets

supershal:
This PR's aim is to make it easy to supply overrides to `konvoy-build` command as we are adding more combinations (OS/FIPS/offline) etc. 
- change `overrides` flag's type to `pflags.StringSlicP` https://pkg.go.dev/github.com/spf13/pflag#StringSliceP which will allow comma separated list of override files. 
- It does not break existing behavior of providing multiple override files using separate `--overrides` flag
- It will easy to specify overrides at runtime in CI job. 
  ex.

```bash
cat <<EOF > release-overrides.yaml
build_name_extra: -release
EOF

make devkit.run WHAT="make ADDITIONAL_OVERRIDES=release-overrides.yaml rhel84-fips-offline"
  ```
This will result in

```bash
./bin/konvoy-image build images/ami/rhel-79.yaml \
-v 6 \
--overrides=overrides/fips.yaml,overrides/offline-fips.yaml,release-overrides.yaml
```

* https://jira.d2iq.com/browse/D2IQ-83343

In future iterations, I wanted to check if its would be useful to support `--extra-vars` flag in `konvoy-build` command which would allow providing values without creating file. ex `--extra-vars fips.enabled=true`.  This flag is already supported in `konvoy-image provision` command. 

```release-note

```


hack/os-packages/Makefile
```diff
@@ -91,7 +91,7 @@ clean-artifacts:
 	rm -rf $(shell pwd)/artifacts
 	for image in $$(\
 		docker image ls \
-			--filter=reference='$(RPM_PACKAGES_GETTER_IMAGE)*' \
+			--filter=reference='$(rpm_packages_getter_image)*' \
```
supershal, line 94: slipping in a fix that broke CI job.

cmd/konvoy-image/cmd/args.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func addOverridesArg(fs *pflag.FlagSet, overrides *[]string) {
-	fs.StringArrayVar(overrides, "overrides", []string{}, "a list of override YAML files")
+	fs.StringSliceVar(overrides, "overrides", []string{}, "a comma separated list of override YAML files")
```
dkoshkin, line 10: So its still possible to set `--overrides=1.yaml --overrides=2.yaml` right? I saw the PR description but was a little confused if this meant only "a comma separated list"

dkoshkin (approved):

cmd/konvoy-image/cmd/args.go
supershal: yes. it is still possible. however I was not sure better way to specify that in short one line in the flag description. So I just mentioned latest way to specify the flag values. any suggestion on the description change? or the current change in description is fine?

faiq:
Great change dude

PR #230, opened 2022-02-06
feat: support azure

jkoelker:

Adds Azure support to KiB

* https://jira.d2iq.com/browse/D2IQ-85255

```release-note
Initial Azure Support
```


jkoelker:
I've build the release artifacts and uploaded them to: https://github.com/jkoelker/konvoy-image-builder/tree/jk/azure_snapshot/dist

Only Centos 7 is currently supported / tested. The easiest way to test this is to have the cli configured, but it also *should* support setting the following environment variables:

```bash
AZURE_LOCATION
AZURE_CLIENT_ID
AZURE_CLIENT_SECRET
AZURE_SUBSCRIPTION_ID
```

I believe there may be dragons with `EnsureImageDescriptions`, and may fail if the existing one already exists. I need to do more testing and make it idempotent. That said I'm not really thrilled with the idea of creating all the Gallery and ImageDescription, but azure images are a little funky in that every actual image is just a version of an existing description.

I'd like some feedback if we should even support creating them if they dont' exist, perhaps just checking that they do exist and failing is the right thing to do. I'm not really sure.

jkoelker:
Welp looks like there is a bit more todo with the config merging.. setting back to draft for now.

jkoelker:
I've build the release artifacts and uploaded them to: https://github.com/jkoelker/konvoy-image-builder/tree/jk/azure_snapshot/dist

Only Centos 7 is currently supported / tested. The follow env vars need to be set (everything except `AZURE_CLIENT_SECRET` can also be specified via CLI or override files):

```bash
AZURE_LOCATION
AZURE_CLIENT_ID
AZURE_CLIENT_SECRET
AZURE_TENANT_ID
AZURE_SUBSCRIPTION_ID
```

I'm not really thrilled with the idea of creating all the Gallery and ImageDescription, but azure images are a little funky in that every actual image is just a version of an existing description, and it requires the description to be created prior to packer running. 

I'd like some feedback if we should even support creating them if they don't exist, perhaps just checking that they do exist and failing is the right thing to do. I'm not really sure.

jkoelker:
Rebased on main to remove last of dependent prs that are now merged.

cmd/konvoy-image-wrapper/cmd/wrapper.go
```diff
@@ -166,6 +169,25 @@ func (r *Runner) setAWSEnv() error {
 	return nil
 }
 
+func (r *Runner) setAzureEnv() error {
```
supershal, line 172: this function never returns any error. The return type should not be `error`

jkoelker: Ah I forgot to have it get called in a loop, I'll remove the error return for now and if we ever another we can refactor it to have a common api for adding the env vars.

supershal (approved):

jkoelker:
Just a rebase to get the changes from #251 to let the OVA offline tests to pass

jkoelker:
Hrm, I'm not sure what the issue is with the 7.9 Offline OVA. It runs but then times out waiting for it to shutdown. I don't think anything in this PR is causing that (the release pr is also failing in the same way), and the 8.4 OVA does not fail. As I don't have vsphere access, @supershal @faiq if either of ya'll have time would you be able to look into this sometime?

If it is not an issue to just ignore it, let me know and I'll merge this in as is.

jkoelker:
Rebased on main, and added the SSH args to ansible and packer sections in `pkg/packer/manifests/azure/packer.json.tmpl` to match aws template key acceptance.

jkoelker:
My Azure trial subscription expired today, so I wasn't able to build it after rebasing, getting the subscription fixed and will build it. I'll hold off on merging until I get a validation run, since we dont' have CI for azure setup yet.

docs/cli/konvoy-image_build_azure.md
```diff
@@ -0,0 +1,52 @@
...
+      --gallery-image-sku string           the gallery image sku to set
+      --gallery-name string                the gallery name to publish the image in (default "konvoy")
+  -h, --help                               help for azure
+      --kubernetes-version string          The version of kubernetes to install. Example: 1.21.6
+      --location string                    the location in which to build the image (default "westus2")
+      --overrides strings                  a comma separated list of override YAML files
+      --packer-manifest string             provide the path to a custom packer manifest
+      --packer-on-error string             [advanced] set error strategy for packer. strategies [cleanup, abort, run-cleanup-provisioner]
+      --packer-path string                 the location of the packer binary (default "packer")
+      --resource-group string              the resource group to create the image in (default "konvoy")
```
joejulian, line 35: Should we be using "dkp" instead of "konvoy"?

pkg/azure/azure.go
```diff
@@ -0,0 +1,231 @@
...
+	galleriesClient := armcompute.NewGalleriesClient(subscriptionID, cred, nil)
+
+	pollerResp, err := galleriesClient.BeginCreateOrUpdate(
+		ctx,
+		description.ResourceGroupName,
+		description.GalleryName,
+		armcompute.Gallery{
+			Location: to.StringPtr(location),
+			Properties: &armcompute.GalleryProperties{
+				Description: to.StringPtr("This is gallery description."),
```
joejulian, line 124: Is this the correct string?

pkg/packer/manifests/azure/packer.json.tmpl
```diff
@@ -0,0 +1,233 @@
...
+    "image_version": "latest",
+    "konvoy_image_builder_version": "0.0.1",
+    "kubernetes_full_version": "",
+    "manifest_output": "manifest.json",
+    "plan_image_offer": "",
+    "plan_image_publisher": "",
+    "plan_image_sku": "",
+    "private_virtual_network_with_public_ip": "",
+    "resource_group_name": "konvoy",
+    "shared_image_gallery_name": "konvoy",
```
joejulian, line 27: Should these be `dkp`?

pkg/azure/azure.go
jkoelker: Lulz, it is _a_ correct string ;). Updated to `DKP Gallery`.

jkoelker:
Updated to switch `konvoy` -> `dkp` in the default created resource names, and rebased on main. Still working on getting my subscription updated to manually test.

PR #231, opened 2022-02-07
fix: configure NetworkManager to prevent interfering with interfaces

supershal:
- Fixes node to pod network for RHEL 8.4. 
RHEL 8.4 comes with NetworkManager v1.32.10 which creates conflicting entry in the routing table that blocks calico from routing node to pod traffic.
This PR adds configuration to prevent NetworkManager interfere with the interfaces as well as disables nm-setup-cloud.timer service which creates conflicting entries in the route table.

The NetworkManager has potential fix (not tested for kubernetes networking) in v1.32.12 . However this rpm version is not available for RHEL 8.X releases.

References: 
- https://github.com/projectcalico/calico/issues/4662
- https://projectcalico.docs.tigera.io/maintenance/troubleshoot/troubleshooting#configure-networkmanager
- https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/740

* https://jira.d2iq.com/browse/D2IQ-83343


supershal:
The one off e2e test that uses the `RHEL 8.4` AMI with this fix passed successfully. https://teamcity.mesosphere.io/viewLog.html?buildId=3345853&buildTypeId=Konvoy2_E2eTestAWSAirGappedFIPS&tab=buildResultsDiv&branch_Konvoy2=%3Cdefault%3E

Makefile
```diff
@@ -6,7 +6,7 @@ OS := $(shell uname -s)
 INTERACTIVE := $(shell [ -t 0 ] && echo 1)
 
 # BUILD_DRY_RUN determines the value of the --dry-run flag of the build command. Should be 'true' or 'false'.
-BUILD_DRY_RUN := true
+BUILD_DRY_RUN ?= true
```
dkoshkin, line 9: Was this not being overridden with `:=`?

ansible/roles/sysprep/tasks/redhat.yml
```diff
@@ -44,6 +44,38 @@
 - name: Reset network interface IDs
   shell: sed -i '/^\(HWADDR\|UUID\)=/d' /etc/sysconfig/network-scripts/ifcfg-*
 
+
+- name: configure NetworkManager on RHEL 8.x
```
dkoshkin, line 48: This is not the correct place to do this, as this only runs when using Packer.
I _think_ we want another role where these, `Disable swap service and ensure it is masked` and `Ensure auditd is started`? are run.

cc @faiq @jkoelker WDYT?

faiq:
Thanks for taking the time to solve it. The tasks look fine to me. Share the same concern about it not working for pre-provisioned hosts

ansible/roles/sysprep/tasks/redhat.yml
supershal: Good point about pre-provisioned. I was wondering what the `sysprep` boolean condition was for.  The `networking` role does not have OS specific tasks now. I can add this rhel 8 specific task there. Also I believe that this task should be removed once RHEL 8.x is able to take the latest NetworkManager rpm.

Makefile
supershal: it may be working. let me test it and confirm.

ansible/roles/sysprep/tasks/redhat.yml
```diff
@@ -44,6 +44,38 @@
...
+  when: ansible_distribution_major_version == '8'
+  block:
+  - name: ensure NetworkManager's conf.d directory exists
+    file:
+      path: /etc/NetworkManager/conf.d
+      state: directory
+
+  # prevent NetworkManager to create conflicting entries in the routing table.
+  # https://projectcalico.docs.tigera.io/maintenance/troubleshoot/troubleshooting#configure-networkmanager
+  - name: configure network manager to prevent it from interfearing with interfaces
```
joejulian, line 58:
```suggestion
  - name: configure network manager to prevent it from interfering with interfaces
```

ansible/roles/networking/tasks/main.yaml
```diff
@@ -1,124 +1,157 @@
-  - name: load br_netfilter kernel module
-    modprobe:
-      name: br_netfilter
-      state: present
-    register: br_netfilter
+- name: load br_netfilter kernel module
```
supershal, line 1: Fixed indentation on this file. somehow it was everything was indented at 4 spaces.

ansible/roles/sysprep/tasks/redhat.yml
supershal: Moved the task to `networking` role.

dkoshkin (approved):

faiq (approved):

PR #233, opened 2022-02-11
fix: create systemd drop-in to make NetworkManager-cloud-setup service no-op

supershal:
- creates drop-in to override for NetworkManager-cloud-setup service (`nm-cloud-setup` ) so that it skips configuring routing table with conflicting entry. The conflicting entry in route table blocks host to container traffic.

**References:**
"Environment Variable" section at: https://developer-old.gnome.org/NetworkManager/stable/nm-cloud-setup.html
RedHat bug: https://bugzilla.redhat.com/show_bug.cgi?id=2007341
NetworkManager issue: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/740
NetworkManager configuration to prevent configuring calico interfaces: https://projectcalico.docs.tigera.io/maintenance/troubleshoot/troubleshooting#configure-networkmanager

Tested the final AMI successfully with e2e tests: https://teamcity.mesosphere.io/viewLog.html?buildId=3349543&tab=buildLog


ansible/roles/networking/tasks/main.yaml
```diff
@@ -57,6 +57,39 @@
...
     - iptables_exists.rc == 0
 
+- name: ensure NetworkManager's conf.d directory exists
+  file:
+    path: /etc/NetworkManager/conf.d
+    state: directory
+
+# prevent NetworkManager to create conflicting entries in the routing table.
+# https://projectcalico.docs.tigera.io/maintenance/troubleshoot/troubleshooting#configure-networkmanager
+- name: configure network manager to prevent it from interfering with calico interfaces
```
supershal, line 67: moved this task out to apply it to all OS family as it is not OS specific according to the docs.

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@
 * rhel82 FIPS ([#200](https://github.com/mesosphere/konvoy-image-builder/issues/200)) ([1688a02](https://github.com/mesosphere/konvoy-image-builder/commit/1688a028fc8a350eb04c3bd1355f40c06b90a2f4))
 
 
-### Bug Fixes
+### Bugfixes
```
supershal, line 16: making linter happy. 
```
/github/workspace/CHANGELOG.md
  16:5   ✓ error  Incorrect usage of the term: “Bug Fixes”, use “Bugfixes” instead  terminology
  30:37  ✓ error  Incorrect usage of the term: “repo”, use “repository” instead     terminology
```
however since this file is auto generated and auto merged, we have to find out if there is any other way to fix/ignore this.

ansible/roles/networking/tasks/main.yaml
```diff
@@ -57,6 +57,40 @@
...
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '8'
+  block:
+  - name: ensure /etc/systemd/system/nm-cloud-setup.service.d exists
+    file:
+      path: /etc/systemd/system/nm-cloud-setup.service.d
+      state: directory
+  - name: create drop-in to disable nm-cloud-setup for AWS
+    copy:
+      dest: /etc/systemd/system/nm-cloud-setup.service.d/override.conf
```
dkoshkin, line 89: Do we not need a `daemon-reload` after?

supershal: My understanding is that running `nm-cloud-setup` service is not needed while building the AMI itself. reloading the daemon or restarting it while building AMI should not matter since no provisioning code making node to container network call that would break. 
However, when the `override.conf` is baked into the AMI and later when this AMI is launched to create k8s node, systemctl will use the `override.conf` from beginning to skip setting up the routing table entry.

dkoshkin: Ahh yeah you're right, thanks for the explanation!

dkoshkin (approved):
I see that konvoy e2e tests worked too https://teamcity.mesosphere.io/buildConfiguration/Konvoy2_E2eTestsAWSAirGappedFIPSRhel84/3349543?buildTab=overview, thanks for running that.

PR #235, opened 2022-02-14
fix(release): run goreleaser on release publish

jkoelker:

Unforts events created by `GITHUB_TOKEN` don't trigger workflows. This means that the `on: push: tags` event was never being sent to trigger the `GoReleaser` workflow.

Instead we can use the output of `release-please` to run the steps needed to fill in the artifacts.

Ref: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
Ref: https://github.com/google-github-actions/release-please-action/issues/400#issuecomment-1025264971

```release-note

```


.github/workflows/release-main.yml
```diff
@@ -3,12 +3,14 @@
...
     branches:
       - main
+
 name: release-please
 jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
-      - uses: GoogleCloudPlatform/release-please-action@v3
+      - uses: google-github-actions/release-please-action@v3
```
jkoelker, line 12: They renamed the repo a few days after we added the workflow :rofl:

dkoshkin (approved):

joejulian (approved):

PR #242, opened 2022-02-18
Revert "build(deps): bump alpine from 3.14.2 to 3.15.0"

faiq:
From #180 

Reverts https://github.com/mesosphere/konvoy-image-builder/pull/148 to fix this issue https://githubmemory.com/repo/atmoz/sftp/issues/296

It appears that the keys packer is creating are not compatible with this new version of open-ssh (8.8) that is shipped with alpine. This is a quick measure until we can figure out how to get packer to create different key types that are compatible with open-ssh 8.8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ ARG BASE=mesosphere/konvoy-image-builder:latest-devkit
 # hadolint ignore=DL3006
 FROM ${BASE} as devkit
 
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):

dkoshkin:
How can we run our e2e tests in the same way that our customers will?

faiq:
I think it would involve building the output container and having the kib builds use that instead of the devkit.run

supershal (approved):

PR #243, opened 2022-02-19
refactor(app): split config handing

jkoelker:
**What problem does this PR solve?**:

* Pulls the config management out into a separate type with helper functions.
* Adds JSON path style access to `Config` type.
* Add unit testing of config management.
* Removes empty packages

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-85254

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #247, opened 2022-02-27
fix(cmd): add subcommads to build and generate

jkoelker:

Deprecates `build` and `generate` defaulting `aws` images, and adds `aws` subcommands to each.

* https://jira.d2iq.com/browse/D2IQ-85253

*NOTE* the `cobra` `Deprecated` field is not set, since that causes the command (and all its subcommands) to not be `Active` and is then excluded from CLI doc generation.

```release-note
CLI: `build` and `generate` commands defaulting to `aws` images is deprecated. Use the `aws` subcommand to those commands to build `aws` images.
```


supershal:
Thank you for putting this much needed refactoring. However I think it would be better if we can discuss UX once with the team. The KIB binary would be included in the unified DKP single cli in future version. So if we are making breaking changes, we can come up with the changes that will align with other subcommands of dkp. 

another thought: All provider specific arguments can also be provided in the `override` or `/image/[platform]/[os-version].yaml` file. what do you think of removing subcommand for the provider itself? I was wondering how do we pick and choose to expose only some provider specific parameters per parameters as flags and not others. OR we thinking to keep only mandatory configuration as `flags` in subcommand? This can leave some configuration in flags and some in override files.

supershal (commented):
Thank you for putting this much needed refactoring. However I think it would be better if we can discuss UX once with the team. The KIB binary would be included in the unified DKP single cli in future version. So if we are making breaking changes, we can come up with the changes that will align with other subcommands of dkp.

another thought: All provider specific arguments can also be provided in the override or /image/[platform]/[os-version].yaml file. what do you think of removing subcommand for the provider itself? I was wondering how do we pick and choose to expose only some provider specific parameters per parameters as flags and not others. OR we thinking to keep only mandatory configuration as flags in subcommand? This can leave some configuration in flags and some in override files.

supershal:
we discuss konvoy improvements in sprint grooming today. We will schedule refactoring stories later to address changes. I saw your azure changes too. with that perspective, splitting to sub command would be beneficial.  
@dkoshkin any thoughts?

jkoelker:
Ah totes, yea the UX totally need some love. Generally though, anything that would be variable (credentials, destination, source ami, etc) I think we'd want to be exposed in the CLI, otherwise scripting becomes harder since the user would have to wrap it a template of some sort to then write out the override yaml for each one.

With this we're maintaining the current "default" behavior of the commands, while also providing the ability to have the multiple provider options like that exposed out so we don't have a bunch of `--aws-*`, `--azure-*`, `--XXX-*` flags added.

jkoelker:
Just a rebase ontop of `main` now that #243 is merged in

supershal (approved):
Thank you for this improvement. We can improve the UX iteratively to support all providers and make it more in line with single dkp cli specs.

PR #271, opened 2022-03-24
fix(ansible): allow rsa public keys

jkoelker:

Looks like we do need both the Host and PublicKey options to enable ssh in ansible from the newer openssh version. (Found this while building an image from main with the wrapper).

```release-note

```


jkoelker:
So because the devkit is debian, and the prod image is apline we have a split in what options are allowed ;(. I know originally the devkit was also alpine, but it got switched before opensourcing it due to `bash` vs `sh`-isms.

What would the feeling be to switch the devkit back to alpine and just make sure to install bash and set it as the default shell? This will make sure we are using the same base platform that the wrapper is, just with go and all the extra tools installed.

Dockerfile.devkit
```diff
@@ -4,19 +4,18 @@ FROM golangci/golangci-lint:v1.44.2-alpine as golangci-lint
 FROM goreleaser/goreleaser:v1.7.0 as goreleaser
 FROM docker:20.10 as docker
 
-FROM golang:1.17.8 as builder
+FROM golang:1.17.8-alpine3.15 as builder
```
faiq, line 7: YES

faiq (approved):
YOU ROCK

dkoshkin (approved):

PR #290, opened 2022-03-31
feat(azure): support ubuntu images

jkoelker:

Adds azure image definitions for ubuntu and additonal python packages.

```release-note

```


ansible/roles/packages/tasks/debian.yaml
```diff
@@ -1,81 +1,68 @@
 ---
-  - name: install apt-transport-https remote deb package
```
jkoelker, line 2: I fixed the indentation in this file as well, so it hit every line ;(

ansible/roles/packages/tasks/debian.yaml
```diff
@@ -1,81 +1,68 @@
...
-    retries: 3
-    delay: 3
+- name: install common debs
+  apt:
+    name:
+      - chrony
+      - libseccomp2
+      - nfs-common
+      - python3-cryptography
+      - python3-pip
```
jkoelker, line 28: I consolidated the different sections into just this one like we do for RPM distros

ansible/roles/providers/tasks/azure-pip.yml
```diff
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+- name: install pip
+  package:
+    name: python3-pip
+    state: present
+  when:
+    - ansible_os_family == "Debian"
```
jkoelker, line 20: This is technically redundant given that `packages` should run first, but it bring in parity with `ansible/roles/providers/tasks/aws.yml` which also has an ensure here.

goss/goss.yaml
```diff
@@ -1,3 +1,11 @@
+{{ $distro := .Vars.OS -}}
+{{ if contains "ubuntu" $distro -}}
+  {{ $distro = "ubuntu" -}}
+{{ end -}}
```
jkoelker, line 4: Unfortuantly the ubuntu `distributions` / offers in azure are kinda cray cray. `0001-com-ubuntu-server-focal-20_04-lts-gen2` and `ubuntuserver-18_04-lts-gen2` so normalizing them here like this to get the goss import to work correctly.

PR #297, opened 2022-04-06
chore(make): reorg build targets and add Azure

jkoelker:

DRY up the current build targets by using `pattern` targets.

The basis is a `build-${provider}-${distro}-${version}` pattern target that will parse out the parts from the target name. Other pattern targets end up calling this target setting the various `ADDITIONAL_OVERRIDES` or `ADDITIONAL_ARGS`. They then provide the following pattern targets:

`${provider}-${distro}-${version}_fips`
`${provider}-${distro}-${version}_offline`
`${provider}-${distro}-${version}_offline-fips`
`${provider}-${distro}-${version}_nvidia`

All the existing targets are now just aliases to the dynamic pattern targets. In the future if we stay consistent with the image name pattern we have for aws/azure/vsphere, we won't have to have any alias patterns and can just setup CI to call the pattern targets directly.

I pulled the images targets out into their own makefile as well. The root one is getting kinda gnarly.

Adds Azure build using these this new target pattern.

*https://jira.d2iq.com/browse/D2IQ-85258

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,7 @@ ifneq ($(shell command -v docker),)
 	endif
 endif
 
+include make/ci.mk
 include make/images.mk
 include hack/pip-packages/Makefile
 include test/infra/aws/Makefile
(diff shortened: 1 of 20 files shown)


supershal (approved):
Thanks for the refactoring code and make it extensible.

PR #298, opened 2022-04-06
build: Attempt to build arm64 images

jimmidyson:

```release-note

```


Dockerfile
```diff
@@ -24,10 +26,10 @@ RUN apk add --no-cache \
     && pip3 install --no-cache-dir --requirement /tmp/requirements.txt \
     && rm -rf /root/.cache
 
-COPY --from=devkit /usr/local/bin/goss /usr/local/bin/
-COPY --from=devkit /usr/local/bin/mindthegap /usr/local/bin/
-COPY --from=devkit /usr/local/bin/packer /usr/local/bin/
-COPY --from=devkit /usr/local/bin/packer-provisioner-goss /usr/local/bin/
+ARG BUILDARCH
+COPY --from=devkit /usr/local/bin/goss-amd64 /usr/local/bin/goss
```
jimmidyson, line 30: Shouldn't this be targetarch too?

Dockerfile
```diff
@@ -27,7 +27,11 @@ RUN apk add --no-cache \
     && rm -rf /root/.cache
 
 ARG BUILDARCH
-COPY --from=devkit /usr/local/bin/goss-amd64 /usr/local/bin/goss
+# we copy this to remote hosts to execute GOSS
```
jimmidyson, line 30: This is really useful explanation! Thank you!

Dockerfile
```diff
@@ -24,10 +26,10 @@ RUN apk add --no-cache \
...
     && rm -rf /root/.cache
 
-COPY --from=devkit /usr/local/bin/goss /usr/local/bin/
-COPY --from=devkit /usr/local/bin/mindthegap /usr/local/bin/
-COPY --from=devkit /usr/local/bin/packer /usr/local/bin/
-COPY --from=devkit /usr/local/bin/packer-provisioner-goss /usr/local/bin/
+ARG BUILDARCH
+COPY --from=devkit /usr/local/bin/goss-amd64 /usr/local/bin/goss
+COPY --from=devkit /usr/local/bin/packer-${BUILDARCH} /usr/local/bin/packer
+COPY --from=devkit /usr/local/bin/packer-provisioner-goss-${BUILDARCH} /usr/local/bin/packer-provisioner-goss
```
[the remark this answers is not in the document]
faiq: BUILDARCH in this case is either amd64 or arm64 isn't it the same thing?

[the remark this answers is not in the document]
jimmidyson: Depends where the binary is running - if it's on the machine where the image is going to run then it will need to be `TARGETARCH`, if it's going to run on the provisioned cluster machines, then it will need to always be `amd64` (until we support arm64 clusters of course...).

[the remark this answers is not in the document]
jimmidyson: This is where i got stuck with this PR - because I don't know where the copied binaries are actually going to run.

faiq (approved):

mkarroqe (approved):
Was able to build successfully on an M1 Mac, LGTM!

Makefile
```diff
@@ -34,8 +34,10 @@ export CGO_ENABLED=0
 # .*    anything after the capture group
 # \1     substitute everything with the 1st capture group
 # p      print it
-export GO_VERSION := $(shell sed -n "s|^go\s*\(\S*\).*|\1|p" go.mod)
+export GO_VERSION := $(shell cat go.mod | grep "go " -m 1 | cut -d " " -f 2)
```
supershal, line 37: much better. :)

pkg/packer/manifests/aws/packer.json.tmpl
```diff
@@ -218,7 +218,7 @@
     },
     {
         "type": "file",
-        "source": "/usr/local/bin/goss",
+        "source": "/usr/local/bin/goss-amd64",
```
supershal, line 221: since the Dockerfile is linking the arch specific binary`ln -s /usr/local/bin/goss-${BUILDARCH} /usr/local/bin/goss`, 
shouldn't we keep it the same path? ` /usr/local/bin/goss`

faiq: we want to upload the specific amd64 because goss is executed on the remote

supershal: 🤦 . Thank you.

Makefile
```diff
@@ -214,23 +219,28 @@ github-token.txt:
...
+
+$(DOCKER_DEVKIT_PHONY_FILE): github-token.txt buildx
 $(DOCKER_DEVKIT_PHONY_FILE): Dockerfile.devkit install-envsubst
 		docker buildx build \
 		$(BUILD_FLAGS) \
-		--tag "$(DOCKER_DEVKIT_IMG)" \
-		--load \
 		$(REPO_ROOT_DIR) \
-	&& touch $(DOCKER_DEVKIT_PHONY_FILE)
+	&& docker load --input /tmp/img.tar && rm /tmp/img.tar && touch $(DOCKER_DEVKIT_PHONY_FILE) && docker images
```
supershal, line 234: Nice. We will have to make similar changes to make it work with goreleaser. I am wondering if we should look into refactor these targets and directly use goreleaser. ironically gorleaser target runs in Devkit container. 🙄

faiq: I think the additional changes in the release-tag action should make it so that each devkit is built beforehand and would use a cache

jimmidyson: Even a multi stage build? This would still benefit from caching.

supershal (approved):
lgtm

PR #287, opened 2022-03-30
fix: fixes version lock to set fact

faiq:
**What problem does this PR solve?**:
fix version lock issue to use a fact instead of a local register

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-NUMBER

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #305, opened 2022-04-12
fix: containerd container lock delete stdout error msg

alejandroEsc:

* https://jira.d2iq.com/browse/D2IQ-87632

```release-note

```

### DIFF
--- a/ansible/roles/containerd/tasks/redhat.yaml
+++ b/ansible/roles/containerd/tasks/redhat.yaml
@@ -9,6 +9,7 @@
     'command_result.stdout is regex(".*versionlock deleted: [1-9]+.*")'
   when:
     - versionlock_plugin_enabled
+    - exportedversionlocklist is defined and exportedversionlocklist.stdout is defined
     - "(
       'containerd.io' in exportedversionlocklist.stdout and
       not 'containerd.io-' + containerd_version in exportedversionlocklist.stdout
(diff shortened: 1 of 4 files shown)


dkoshkin:
Are there other instances that needs to be fixed in the same way? e.g. `kubeadm` https://github.com/mesosphere/konvoy-image-builder/pull/287/files#diff-5f8572a1492839c754b58fc939b8c0ab5a0427a880a69e766a37cdb266ed0338R13

alejandroEsc:
> Are there other instances that needs to be fixed in the same way? e.g. `kubeadm` https://github.com/mesosphere/konvoy-image-builder/pull/287/files#diff-5f8572a1492839c754b58fc939b8c0ab5a0427a880a69e766a37cdb266ed0338R13

good point, i havent checked, will do. Also is it ok to skip this step if stdout is not available? I would guess yes but I would like a confirmation.

alejandroEsc:
@faiq I have a test and golint error out are these expected?

faiq:
So I took a look at the COPS issue linked and it looks like there was an issue with KIB v1.9.1 and version lock the actual issue should have been resolved with the following commit https://github.com/mesosphere/konvoy-image-builder/commit/ebaba83028e79adf8fb295b26961a18a15e9be50 


This error message occurs during the first run of KIB on a machine because nothing is in the versionlock and it warns about it. 

It looks safe to merge to me to suppress the output, but in my opinion the COPS is no longer valid.

Regarding the failing OVA test, that happens on delete. It is unrelated. 

The lint failure is due to some bad config.

Reach out on slack if you need some more clarification on this.

alejandroEsc:
> Regarding the failing OVA test, that happens on delete. It is unrelated.

Looks like that PR fixes these issues already. Feel free to close this and update the JIra ticket that you have a solution already. Do you need reviewers?

faiq:
Closing after conversation.

PR #306, opened 2022-04-12
build: warn the developer if BUILD_DRY_RUN is true

joejulian:
... also linter fix

Warn about a hidden undocumented environment variable.

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ INTERACTIVE := $(shell [ -t 0 ] && echo 1)
 
 # BUILD_DRY_RUN determines the value of the --dry-run flag of the build command. Should be 'true' or 'false'.
 BUILD_DRY_RUN ?= true
+ifeq ($(BUILD_DRY_RUN),true)
+$(warning Warning: BUILD_DRY_RUN is true)
+endif
 
 root_mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 export REPO_ROOT_DIR := $(dir $(root_mkfile_path))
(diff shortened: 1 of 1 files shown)


faiq (approved):

dlipovetsky:
Seems odd to have a warning about the default value. Maybe we were wrong to make it the default?

Or maybe the problem is that we're using make as a CLI framework, but only env vars, and no flags? :laughing:

joejulian:
Maybe so. I don't know why it exists and why it has that default. I ran a make command twice and couldn't see why it wasn't creating an ami. I asked @faiq if it worked for him, and he told me about this variable. I'm trying to add this warning so that whatever reason it exists, any new developers that don't already know that secret will at least have some way of learning about it.

Makefile
```diff
@@ -7,6 +7,9 @@ INTERACTIVE := $(shell [ -t 0 ] && echo 1)
 
 # BUILD_DRY_RUN determines the value of the --dry-run flag of the build command. Should be 'true' or 'false'.
```
dlipovetsky, line 8:
```suggestion
# BUILD_DRY_RUN enables the build command's dry-run mode. In this mode, the build does not produce any permanent artifacts, e.g., an AMI. Should be 'true' or 'false'.
```

dlipovetsky:
> Maybe so. I don't know why it exists and why it has that default. I ran a make command twice and couldn't see why it wasn't creating an ami. I asked @faiq if it worked for him, and he told me about this variable. I'm trying to add this warning so that whatever reason it exists, any new developers that don't already know that secret will at least have some way of learning about it.

Which make target did you run? Maybe that target should disable dry-run?

We added dry-run (and made it the default) in order to avoid creating permanent artifacts as the result of running e2e tests of the build command.

Makefile
joejulian: It's not about the comment, it's that I wouldn't have even guessed that I would have to read the Makefile to understand this variable even exists. With this warning, it gives that hint.

joejulian:
> > Maybe so. I don't know why it exists and why it has that default. I ran a make command twice and couldn't see why it wasn't creating an ami. I asked @faiq if it worked for him, and he told me about this variable. I'm trying to add this warning so that whatever reason it exists, any new developers that don't already know that secret will at least have some way of learning about it.
> 
> Which make target did you run? Maybe that target should disable dry-run?
> 
> We added dry-run (and made it the default) in order to avoid creating permanent artifacts as the result of running e2e tests of the build command.

That makes total sense and for that reason, I think the default should remain. As a developer, it's probably more common that you'll want to test the build process without creating the ami artifact.

Makefile
dlipovetsky: Good point. Would you like to move my text here to the warning you added?

dlipovetsky (approved):

PR #321, opened 2022-04-28
chore(devkit): fix python packages

jkoelker:

The devkit is currently failing to build due to conflicting versions of `psutil` being installed vs needed (`alpine` has 5.8 and `azure-cli` requires 5.9).

To remedy this, move the requirements and pining of what we actually care about into `requirements.txt` and `requirements-devkit.txt`. This allows pip to satisfy all the requirements as binary wheels currently, and will allow us to use dependabot in the future to manage bumping the versions. 

```release-note

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,10 @@ FROM ${BASE} as devkit
 
 FROM alpine:3.15.2
 
-ARG ANSIBLE_VERSION=2.10.7
 ENV ANSIBLE_PATH=/usr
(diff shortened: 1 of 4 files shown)


supershal (approved):

alejandroEsc (approved):
lgtm, is the failing test expected?

supershal:
yes @alejandroEsc . for some reason, the RHEL 7.9 VM is not shutting down. We need to investigate the issue with it.

PR #327, opened 2022-05-08
feat(oracle): support installing RHCK kernel 

jkoelker:

Adds the `oracle_kernel` varaible for ansible. When set to the string `RHCK` it will install the RHCK kernel and set it as the default for the next boot.

* https://jira.d2iq.com/browse/D2IQ-88561

```release-note

```

### DIFF
--- a/ansible/roles/packages/tasks/oracle.yaml
+++ b/ansible/roles/packages/tasks/oracle.yaml
@@ -1,0 +1,76 @@
+---
+
+- name: RHCK kernel
+  when:
+    - oracle_kernel == "RHCK"
+  block:
+    - name: install RHCK kernel
+      yum:
+        name: kernel
+        state: present
+        update_cache: true
+        enablerepo: "{{ 'offline' if offline_mode_enabled else '' }}"
+        disablerepo: "{{ '*' if offline_mode_enabled else '' }}"
+      register: rhck_installation_rpm
+      until: rhck_installation_rpm is success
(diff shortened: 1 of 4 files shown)


jkoelker:
Just a rebase on `main` to keep the CI #fresh

jkoelker:
The CI failures are all the expected ones. The OVA offlines are timing out shutting down, and the CentOS 7 and Azure Ubuntu Nvidia are still WIP.

alejandroEsc (approved):
I have limited ability to understand some of this but it looks like it doing things correctly, I also see that there is some refactoring.

ansible/roles/packages/tasks/oracle.yaml
```diff
@@ -0,0 +1,76 @@
+---
+
```
msdolbey, line 2: I really don't know all the yamls in the world, and this probably doesn't really matter at all - but should there be this extra line here?

jkoelker:
> Do we need an E2E variant for this?

I would like one, but it requires running the image to actually check.

I attempted to add a goss step to check the grub entry, but it was turning into more hassle to shovel the flag around and then goss only checked the running kernel, not the configured grub entry.

I suppose we could add a build to make sure that the ansible doesn't fail, but it's only 1/2 the story since it may succeed in installing it and adding it, but the grubenv / defaults gets set later on. I guess its better than nothing. I'll add that this weekend.

PR #329, opened 2022-05-10
feat: install Containerd from a tar instead of OS packages

dkoshkin:
Changing the way Containerd is installed, where its installed from a tar.
The tar is bundled in the KIB image, this makes version skew and air-gapped much simpler as now users will only need to pass the Kubernetes version and the matching k8s OS package bundle and KIB will have the required Containerd version baked in. 

* https://jira.d2iq.com/browse/D2IQ-89325

Added a `do not merge` as we do not have fips tars yet.

```release-note

```


Dockerfile.devkit
```diff
@@ -107,6 +107,10 @@ COPY --from=docker /usr/local/bin/docker /usr/local/bin/
 COPY --from=govc /govc /usr/local/bin/
 COPY --from=builder /tools /usr/local/bin
 
+ARG CONTAINERD_VERSION
+ADD "https://github.com/containerd/containerd/releases/download/v${CONTAINERD_VERSION}/cri-containerd-cni-${CONTAINERD_VERSION}-linux-amd64.tar.gz" /opt/dkp/containerd.tar.gz
+# FIXME(dkoshkin) use a real FIPS version
+ADD "https://github.com/containerd/containerd/releases/download/v${CONTAINERD_VERSION}/cri-containerd-cni-${CONTAINERD_VERSION}-linux-amd64.tar.gz" /opt/dkp/containerd-fips.tar.gz
```
jkoelker, line 113: Is there a way we can do this all in ansible where its either a URL or if `offline` in the `/artifacts` dir?

dkoshkin: We can and thats how I started in https://github.com/mesosphere/konvoy-image-builder/pull/329/commits/3acc99e30198bec90c6f5163307bd3264db30736

But I've been thinking about this, currently we install the Containerd package from the same offline repo that has Kubernetes packages. This basically couples KIB to a specific version of Kubernetes, Containerd, `dkp` since all these things need to align to work.
Moving this responsibility to KIB greatly simplifies all of it as now users would only need to worry about the Kubernetes version and a version of KIB should be able to work with any combination of Kubernetes and `dkp`.

jkoelker: Hrm, I see, the hozer though is that since the version is still an ansible variable, it can be overridden anyway, and then it will fail because the file isn't baked into the docker layer, and this inflates the size of the docker image by ~50MB.

I totally agree that KIB should be responsible for the pairing, I just think that the ansible portion should be the one doing that, since at the end of the day, packer is just invoking ansible to actually do the provisioning. I dont' think we should be coupling to the use of a particular docker image. In theory there's no reason not to be able to use any devkit version since all its for is making sure packer, ansible, and goss are installed in an "easy" way.

Not a blocker just my $0.02 ;)

jkoelker (approved):

Dockerfile.devkit
dkoshkin: Talking about this in standup, this approach may not be good as we are going to have multiple different tars for each OS/version/FIPS combination.

supershal (approved):
Thanks for running e2e test and validating the PR

ansible/roles/containerd/tasks/install.yaml
```diff
@@ -0,0 +1,55 @@
...
+- name: delete /opt/cni directory
+  file:
+    path: /opt/cni
+    state: absent
+
+# Remove /etc/cni directory, as we will configure cni later
+- name: delete /etc/cni directory
+  file:
+    path: /etc/cni
+    state: absent
```
dlipovetsky, line 43: Our containerd tarballs will not include CNI. We can remove these workarounds in the future.

dlipovetsky:
@dkoshkin Here's what's included in the containerd release tarball that upstream image-builder downloads:

```txt
cri-containerd-cni-1.6.4-linux-amd64/
├── etc
│   ├── cni
│   │   └── net.d
│   │       └── 10-containerd-net.conflist
│   ├── crictl.yaml
│   └── systemd
│       └── system
│           └── containerd.service
├── opt
│   ├── cni
│   │   └── bin
│   │       ├── bandwidth
│   │       ├── bridge
│   │       ├── dhcp
│   │       ├── firewall
│   │       ├── host-device
│   │       ├── host-local
│   │       ├── ipvlan
│   │       ├── loopback
│   │       ├── macvlan
│   │       ├── portmap
│   │       ├── ptp
│   │       ├── sbr
│   │       ├── static
│   │       ├── tuning
│   │       ├── vlan
│   │       └── vrf
│   └── containerd
│       └── cluster
│           ├── gce
│           │   ├── cloud-init
│           │   │   ├── master.yaml
│           │   │   └── node.yaml
│           │   ├── cni.template
│           │   ├── configure.sh
│           │   └── env
│           └── version
└── usr
    └── local
        ├── bin
        │   ├── containerd
        │   ├── containerd-shim
        │   ├── containerd-shim-runc-v1
        │   ├── containerd-shim-runc-v2
        │   ├── containerd-stress
        │   ├── crictl
        │   ├── critest
        │   ├── ctd-decoder
        │   └── ctr
        └── sbin
            └── runc
```

Here's what I think we should keep for _our_ containerd tarballs:

```txt
konoy-cri-containerd-1.6.4-linux-amd64/
├── etc
│   ├── crictl.yaml
│   └── systemd
│       └── system
│           └── containerd.service
└── usr
    └── local
        ├── bin
        │   ├── containerd
        │   ├── containerd-shim
        │   ├── containerd-shim-runc-v1
        │   ├── containerd-shim-runc-v2
        │   ├── containerd-stress
        │   ├── crictl
        │   ├── critest
        │   ├── ctd-decoder
        │   └── ctr
        └── sbin
            └── runc
```

I'm not sure we even need all of these. I think minimal set is:

```txt
konoy-cri-containerd-1.6.4-linux-amd64/
├── etc
│   ├── crictl.yaml
│   └── systemd
│       └── system
│           └── containerd.service
└── usr
    └── local
        ├── bin
        │   ├── containerd
        │   ├── containerd-shim
        │   ├── containerd-shim-runc-v1
        │   ├── containerd-shim-runc-v2
        │   ├── crictl
        │   └── ctr
        └── sbin
            └── runc
```

faiq (approved):

ansible/group_vars/all/defaults.yaml
```diff
@@ -32,9 +30,8 @@ aws_images: []
 
 extra_images: []
 
-# https://github.com/containerd/containerd/releases/download/v1.4.11/cri-containerd-cni-1.4.11-linux-amd64.tar.gz
-containerd_url: https://github.com/containerd/containerd/releases/download/v{{ containerd_version }}/cri-containerd-cni-{{ containerd_version }}-linux-amd64.tar.gz
-containerd_sha256: 80c47ec5ce2cd91a15204b5f5b534892ca653e75f3fba0c451ca326bca45fb00
+# FIXME (dkoshkin) use an OS specific and FIPS tar
```
faiq, line 33:
```suggestion
```

ansible/roles/containerd/tasks/install.yaml
dlipovetsky: Removed

dlipovetsky:
Squashed fixups, reworded commits, and force-pushed. Side benefit: previously failed tests run again.

PR #330, opened 2022-05-11
fix: adds a v1 to copy path

faiq:
Adds a v1 that now get appended during tagged builds which goreleaser runs. The normal builds seem to not use goreleaser so tests should not be affected.

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -332,7 +332,7 @@ build.snapshot:
 	#                copying the dist folder into the temporary folder
 	#                that it uses as its docker build context ;(.
 	mkdir -p bin
-	cp dist/konvoy-image_linux_amd64/konvoy-image bin/konvoy-image
+	cp dist/konvoy-image_linux_amd64_v1/konvoy-image bin/konvoy-image
 	goreleaser --parallelism=1 --skip-publish --snapshot --rm-dist
 
 .PHONY: diff


dkoshkin:
Do you know where `v1` comes from?

joejulian (approved):

supershal (approved):

jkoelker (approved):
You'll want to rebase on main to pick up https://github.com/mesosphere/konvoy-image-builder/commit/1087843fedcdb09e285b71b9eec6e9f99da5f7d3 which will fix the build

joejulian:
> Do you know where `v1` comes from?

goreleaser. They added the ability to have multiple release versions. An excerpt from the documentation:

from the `builds` section
```
    # GOAMD64 to build when GOARCH is amd64.
    # For more info refer to: https://golang.org/doc/install/source#environment
    # Default is only v1.
    goamd64:
      - v2
      - v3
```

supershal:
If you dont specify in the goreleaser config, it will be default to `v1`. so we are good for now. but if we decided to change it in future, we can create array of the goamd64 values 
in the config: https://github.com/goreleaser/goreleaser/blob/main/internal/builders/golang/build.go#L67

joejulian:
rebased

joejulian:
Can some supreme overlord please admin merge this tiny fix for us plebes? The only reason those two required tests failed is that the SLES repo mirror that was randomly selected was broken.

PR #336, opened 2022-05-12
fix: remove extra "release"  keyword from vsphere template name

supershal:
This is one of the first PR to help cleanup vSphere environment and automate it.
- Removes extra release keyword from the name of the template: `konvoy-ova-vsphere-rhel-79-release-release-1.22.8+fips.0-1652280664`
- Removes `ova-vsphere` keyword to allow more room in the VM name. The VM name length is limited. 
  We are planning to add KIB version in the name of the release template so that we can automate "lookup" functionality when running e2e tests. This will help make image lookup similar as we do in AWS AMI lookup and avoid creating a PR to change the template name.
- The TC job to create template will add KIB version in the template name. ex. `buildname_extra: release-v1.13.2`

I will file another PR that will cleanup old released template from vSphere.

### DIFF
--- a/pkg/packer/manifests/vsphere/packer.json.tmpl
+++ b/pkg/packer/manifests/vsphere/packer.json.tmpl
@@ -2,7 +2,7 @@
     "variables": {
         "ansible_extra_vars": "",
         "build_timestamp": "{{timestamp}}",
-        "vm_name": "konvoy-ova-{{user `build_name`}}{{user `build_name_extra`}}-{{user `kubernetes_full_version` }}-{{user `build_timestamp`}}",
+        "vm_name": "konvoy-{{user `build_name`}}-{{user `kubernetes_full_version` }}-{{user `build_timestamp`}}",
         "cluster": "",
         "datastore": "",
         "folder": "",
--- a/images/ova/rhel-79.yaml
+++ b/images/ova/rhel-79.yaml
@@ -1,6 +1,6 @@
 ---
 download_images: true
-build_name: "vsphere-rhel-79"
+build_name: "rhel-79"
 packer_builder_type: "vsphere"
 guestinfo_datasource_slug: "https://raw.githubusercontent.com/vmware/cloud-init-vmware-guestinfo"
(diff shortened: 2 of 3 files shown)


pkg/packer/manifests/vsphere/packer.json.tmpl
```diff
@@ -2,7 +2,7 @@
     "variables": {
         "ansible_extra_vars": "",
         "build_timestamp": "{{timestamp}}",
-        "vm_name": "konvoy-ova-{{user `build_name`}}{{user `build_name_extra`}}-{{user `kubernetes_full_version` }}-{{user `build_timestamp`}}",
```
supershal, line 5: build_name_extra is already added by the go code. so removing repeatation here.  https://github.com/mesosphere/konvoy-image-builder/blob/main/pkg/app/config.go#L195

dkoshkin: What will the resulting name be? Doesn't the `-release` come from `build_name_extra` which are we now removing?

supershal: I am planning to change the TC job to add the branch_name in the `build_name_extra` 
```
cat <<EOF > release-overrides.yaml
build_name_extra: release-%teamcity.build.branch%
EOF
```
the go code adds the `build_name_extra` so we dont need to add it explicitly in packer template.
https://github.com/mesosphere/konvoy-image-builder/blob/main/pkg/app/config.go#L98

The resulting template name will look like: `konvoy-ova-vsphere-rhel-79-release-v1.13.2-1.22.8+fips.0-1652304962`

dkoshkin: Thanks 👍,
[I see its not in AMI either](https://github.com/mesosphere/konvoy-image-builder/blob/90c84fc9d9403e150bf1075c277f17c1f5e33ce7/pkg/packer/manifests/aws/packer.json.tmpl#L60)

dkoshkin (approved):

faiq (approved):

PR #350, opened 2022-05-20
feat: create centos images for gcp 

alejandroEsc:
Second part 

* https://jira.d2iq.com/browse/D2IQ-88880

```release-note

```


images/gcp/centos-7.yaml
```diff
@@ -0,0 +1,20 @@
+---
+download_images: true
+
+packer:
+  # The project ID that will be used to launch instances and store images.
+  project_id: "eng-ksphere-platform"
+  # The source image to use to create the new image from.
+  source_image: "centos-7-v20220519"
+  # The source image family to use to create the new image from.
+  source_image_family: "centos-7"
```
faiq, line 10: Please specify 7.9 and not just centos-7

alejandroEsc: Im not sure how to do so for gcp, but will search for it.

ansible/roles/providers/tasks/googlecompute.yml
```diff
@@ -40,3 +40,51 @@
...
+  apt:
+    name: "{{ packages }}"
+    state: present
+    force_apt_get: true
+  vars:
+    packages:
+      - cloud-init
+      - cloud-guest-utils
+      - cloud-initramfs-copymods
+      - cloud-initramfs-dyn-netconf
```
faiq, line 54: cloud-init is a requirement for CAPA. Is there any way we can consolidate this stuff to avoid drift between cloud providers

ansible/roles/providers/tasks/googlecompute.yml
```diff
@@ -40,3 +40,51 @@
...
+      - cloud-init
+      - cloud-init-guestinfo
+  when: ansible_os_family == "Suse"
+
+
+# - name: Unlock password
+#   replace:
+#     path: /etc/cloud/cloud.cfg
+#     regexp: '(?i)lock_passwd: True'
+#     replace: 'lock_passwd: False'
```
faiq, line 83:
```suggestion
```

images/gcp/centos-7.yaml
alejandroEsc: FYI: CentOS Linux release 7.9.2009 (Core)

ansible/roles/providers/tasks/misc.yml
```diff
@@ -86,6 +86,7 @@
     name: chronyd
   when: (packer_builder_type.startswith('amazon') or
     packer_builder_type.startswith('azure') or
+    packer_builder_type.startswith('gcp') or
```
supershal, line 89:
```suggestion
    packer_builder_type.startswith('googlecompute') or
```

Makefile
```diff
@@ -54,6 +54,10 @@ export DOCKER_DEVKIT_AWS_ARGS ?= \
 	--env AWS_DEFAULT_REGION \
 	--volume "$(HOME)/.aws":"/home/$(USER_NAME)/.aws"
 
+export DOCKER_DEVKIT_GCP_ARGS ?= \
+	--env GOOGLE_APPLICATION_CREDENTIALS \
+	--volume "$(HOME)/.gcp":"/home/$(USER_NAME)/.gcp"
```
supershal, line 59: do we need to mount `.gcp` if `GOOGLE_APPLICATION_CREDENTIALS` is exported?

supershal (changes requested):

Makefile
alejandroEsc: `GOOGLE_APPLICATION_CREDENTIALS` is just a pointer to a file that exist in your system. It has to be mounted.

ansible/roles/providers/tasks/googlecompute.yml
alejandroEsc: It seems like image-builder still expected that cloud-init be installed, specially for ubuntu distros. There are checks for these also so I assume that the assumption is that this still exists.

images/gcp/centos-7.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
faiq, line 1: would you mind changing the file to centos79.yaml? 

 
I'm trying to be specific in the point versions we support and I think it's a good practice to get into.

alejandroEsc: This breaks existing files. I would prefer if we do this in another PR and just settle this holistically and not for just this one case.

ansible/roles/providers/tasks/googlecompute.yml
faiq: I see. The point i was making is that `providers/tasks/vsphere.yaml` does something very similar in installing packages and was wondering if we could consolidate those installs in one file.

images/gcp/centos-7.yaml
alejandroEsc: Ha i havent pulled those changes. Alright then no problem.

faiq (approved):
Thank you for changing the name + refactors for options

pkg/packer/manifests/gcp/packer.json.tmpl
```diff
@@ -0,0 +1,203 @@
...
+         "image_builder_version": "{{ user `konvoy_image_builder_version` | clean_resource_name }}",
+         "kubernetes_cni_version": "{{ user `kubernetes_cni_version` }}",
+         "kubernetes_version": "{{ user `kubernetes_full_version` | clean_resource_name }}"
+       },
+       "metadata": {
+
+       },
+       "network": "{{ user `network` }}",
+       "tags": [
+         "alejandro-network-allow-ssh"
```
supershal, line 53: please remove hardcoded tag and use variables.

pkg/packer/manifests/gcp/packer.json.tmpl
```diff
@@ -0,0 +1,203 @@
...
+    "konvoy_image_builder_version": "0.0.1",
+    "kubernetes_full_version": "",
+    "manifest_output": "manifest.json",
+    "network": "",
+    "project_id": "",
+    "zone": ""
+  },
+  "builders": [
+    {
+       "image_name": "{{ (user `distribution`) | lower }}-{{ user `distribution_version` }}{{ user `build_name_extra` }}",
```
supershal, line 20: The `image_name` should be similar in format as we use in AWS where `kuberentes_version` is used.
ex. https://github.com/mesosphere/konvoy-image-builder/blob/main/pkg/packer/manifests/aws/packer.json.tmpl#L63

I think you can use this value for the `"name"` config. 
```
"name": "{{ (user `distribution`) | lower }}-{{ user `distribution_version` }}{{ user `build_name_extra` }}",
```

supershal (changes requested):

pkg/packer/manifests/gcp/packer.json.tmpl
alejandroEsc: oh thank you for pointing that out!

pkg/packer/manifests/gcp/packer.json.tmpl
alejandroEsc: So your suggestion is what we are using for image name. Im not sure I understand, azure is doing the same for the image name and I am following that convention thought I guess that is open to debate. Do you want to also include a timestamp in the name?

supershal: There are configs:
- `name`: Name of the packer build that shows up on the packer output.
- `image_name` (gcp) | `ami_name`(aws) | `managed_image_name` (azure) | `vm_name` (vsphere): The name of the resulted image that would be stored on the cloud provider which will be used to create cluster.

we should be wanting to  include `kubernetes_version` in the image name for all providers. This will help extend automated build chains to `lookup` / `search` images based on the kubernetes version, base os and timestamp. 
The current automated build chain (k8s upstream -> rpms -> KIB -> Konvoy -> dkp) is built for AWS but this will help it to extend other providers too. The build chains are based on kubernetes version so I think the KIB image name should include k8s version to reflect that. 

I think our Azure packer config should be setup similar way too. by just reading through packer documentation, it seems like Azure should be setting proper `managed_image_name` for https://www.packer.io/plugins/builders/azure/arm#managed_image_name-1 and use similar format as AWS AMI: `"konvoy-ami-{{user `build_name`}}-{{user `kubernetes_full_version` | clean_resource_name}}-{{user `build_timestamp`}}"`

alejandroEsc: I understand we may want to do this. However, nothing in the documentation says it is required, and even in our code we do not require this nor do we ask that the user use the mananged name in a certain format. I think you bring up some interesting points, and maybe we should reconsider these pieces as we move forward. I would be ok to change the name now, but I would rather have a conversation around this outside of this PR and then see how we proceed. If we do decide that we want a specific name, then we probably should have a default value for say gallery_image_name, which I cannot find one in the current code.

supershal (approved):

alejandroEsc:
all required passed. I will be merging this now.

PR #367, opened 2022-05-31
fix: correctly set default pause image

dkoshkin:
The default `pause` from Containerd was not getting pulled, causing air-gapped tests to fail.
This is because the task to run get the image relies on `containerd` but it was not in the PATH 

See this example where the output is empty https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_KonvoyImageBuilder_E2eCentOS7/3471607?buildTab=log&focusLine=1183&linesState=491

With the fix - https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_KonvoyImageBuilder_E2eCentOS7/3471618?buildTab=log&focusLine=1188&linesState=495

---

Alternatively maybe we should just add the binaries to `/usr/bin` instead of `/usr/local/bin`?

```release-note

```

### DIFF
--- a/ansible/roles/containerd/tasks/install.yaml
+++ b/ansible/roles/containerd/tasks/install.yaml
@@ -43,3 +43,4 @@
     force: yes
   loop:
     - ctr
+    - containerd
--- a/ansible/roles/containerd/tasks/install.yaml
+++ b/ansible/roles/containerd/tasks/install.yaml
@@ -31,9 +31,9 @@
     extra_opts:
       - --no-overwrite-dir
 
-# We run ctr. When we do, it must be in the PATH of the ansible process. On some distributions, /usr/local/bin is not in
-# the PATH. The symlink makes ctr available in /usr/bin, which is in the PATH across all distributions where we install
-# ctr. We do not install ctr on flatcar, but there it is already in the PATH.
+# Some ansible tasks run ctr and containerd. For these tasks, the executable must be in the PATH of the ansible process.
+# On some distributions, /usr/local/bin is not in the PATH. The symlink makes the executables available in /usr/bin,
+# which is in the PATH across all supported distributions. On flatcar, the executables are already in the PATH.
 - name: symlink ctr to /usr/local/bin
   file:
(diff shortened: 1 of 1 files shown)


dlipovetsky:
> Alternatively maybe we should just add the binaries to /usr/bin instead of /usr/local/bin?

The convention is for non-system-packages to go to `/usr/local`, or `/opt`.

We might be able to fix the `PATH`.

dlipovetsky:
The e2e test passed. It looks like konvoy-image-builder takes a "best effort" approach to pulling the pause image. That seems like a bug that needs to be fixed.

ansible/roles/containerd/tasks/install.yaml
```diff
@@ -43,3 +43,4 @@
     force: yes
   loop:
     - ctr
+    - containerd
```
supershal, line 46: LGTM.  minor suggestion: The task name and comment description is specific to `ctr`. can add info about `containerd`

supershal (approved):

dlipovetsky:
The [task](https://github.com/mesosphere/konvoy-image-builder/blob/8d9c8e924906cf55ddbab161b410fe60eee0d804/ansible/roles/version/tasks/main.yaml#L38-L42) uses the `shell` module

```yaml
- name: determine containerd pause image
  shell: >
    containerd config default |
    sed -n 's/^.*sandbox_image\s*=\s"\(\S*\)".*$/\1/p'
  register: containerd_sandbox_image
```

> runs the command through a shell (/bin/sh) on the remote node.
> -- https://docs.ansible.com/ansible/2.9/modules/shell_module.html

Most likely, the shell is `bash`. Since we're using a pipeline, and the command that fails is _not last command_ in the pipeline, the error is not propagated. We'd need to enable the `pipefail` shell option. This is a classic ansible+shell pitfall. Let me verify that enabling this option does indeed fix the issue here (Although I'd love to find an alternative to the pipeline, honestly.)

dlipovetsky (approved):

PR #369, opened 2022-05-31
fix: Catch errors when the shell module executes a pipeline

dlipovetsky:
Errors are not being caught when the shell module executes a pipeline. For example, see https://github.com/mesosphere/konvoy-image-builder/pull/367#issuecomment-1142351909

```release-note

```


dlipovetsky:
After #367 merged, I rebased on main and force-pushed.

dkoshkin (approved):
Thanks for fixing all tasks

supershal (approved):

dlipovetsky:
> Ubuntu 18.04 and 20.04 tests are failing apparently because `/bin/sh` is not `bash`, and `set -o pipefail` is unrecognized. Will fix.

I tried to fix this by relying on `/usr/bin/env bash`, but that proved tricky to do. So instead I went all-in on the "native" parsing approach introduced recently by @faiq.

dlipovetsky:
Sadly and unexpectedly, the cli_parse module didn't work for my needs, and its errors were inscrutable. So I went with regex parsing.

PR #373, opened 2022-06-01
feat: install googlecompute packer plugin

supershal:
- bakes googlecompute packer plugin in Devkit and KIB wrapper docker images

* https://jira.d2iq.com/browse/D2IQ-90153

```release-note

```


supershal:
Manually tested image creation. Teamcity need to setup with GCP credentials to run it with automation.
```make build-image \
	PROVIDER=gcp \
	ADDITIONAL_OVERRIDES= \
	ADDITIONAL_ARGS="" \
	IMAGE=images/gcp/centos-79.yaml \
	VERBOSITY=0 \
	BUILD_DRY_RUN=true
make[2]: Entering directory '/kib'
make/images.mk:8: Warning: BUILD_DRY_RUN is true
Executing target: bin/konvoy-image
go build \
	-ldflags="-X github.com/mesosphere/konvoy-image-builder/pkg/version.version=v1.16.0-4-gc2b0a1cd5ac6" \
	-o ./bin/konvoy-image ./cmd/konvoy-image/main.go
./bin/konvoy-image build gcp images/gcp/centos-79.yaml \
--dry-run=true \
-v 0 \
 \
 \

2022/06/01 20:18:32 writing new packer configuration to work/centos-7-1654114712-dfdOy
2022/06/01 20:18:32 starting packer build
googlecompute: output will be in this color.

==> googlecompute: Checking image does not exist...
==> googlecompute: Creating temporary RSA SSH key for instance...
==> googlecompute: Using image: centos-7-v20220519
==> googlecompute: Creating instance...
    googlecompute: Loading zone: us-west1-a
    googlecompute: Loading machine type: n1-standard-1
    googlecompute: Requesting instance creation...
    googlecompute: Waiting for creation operation to complete...
    googlecompute: Instance has been created!
    googlecompute: Waiting 20s before adding SSH keys...
==> googlecompute: Waiting for the instance to become running...
    googlecompute: IP: 35.233.149.210
==> googlecompute: Using SSH communicator to connect: 35.233.149.210
==> googlecompute: Waiting for SSH to become available...
==> googlecompute: Connected to SSH!
==> googlecompute: Provisioning with shell script: /tmp/packer-shell2217156501
...
...
````

alejandroEsc:
So i understand the dev-kit pieces. Can we also have either a make entry to download the plugin for us or have it installed ourselves locally for dev?

alejandroEsc:
> I have not been able to use KIB without devkit container when testing locally. I think consistent approach would be to test it through devkit container only. for ex: we install and use goss and some ansible plugins in devkit only.

I dont agree necessarily but I understand. Im ok with merging this, my thoughts on these situations is that if we ask customers to use a wrapper, maybe we should be testing this way too. Anyway lgtm

alejandroEsc (approved):
lgtm

faiq:
> I dont agree necessarily but I understand. Im ok with merging this, my thoughts on these situations is that if we ask customers to use a wrapper, maybe we should be testing this way too. Anyway lgtm 

Totally agree. We are shipping this as a wrapper by default, so the testing and shipping are the same.

supershal:
> I dont agree necessarily but I understand. Im ok with merging this, my thoughts on these situations is that if we ask customers to use a wrapper, maybe we should be testing this way too. Anyway lgtm.

The wrapper is covered by https://github.com/mesosphere/konvoy-image-builder/pull/373/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R29 .
+1. we should be using wrapper in our e2e test.

PR #375, opened 2022-06-01
chore: skip gcp image creation with --dry-run=true

supershal:
- adds default disk size to 80 GB to match with other providers
- skip image creation with dry-run is true
- add dev doc for GCP

```release-note

```

### DIFF
--- a/pkg/packer/manifests/gcp/packer.json.tmpl
+++ b/pkg/packer/manifests/gcp/packer.json.tmpl
@@ -13,7 +13,9 @@
     "manifest_output": "manifest.json",
     "network": "",
     "project_id": "",
-    "zone": ""
+    "zone": "",
+    "disk_size": "80",
+    "disk_type": "pd-standard"
   },
   "builders": [
     {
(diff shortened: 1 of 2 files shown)


pkg/packer/manifests/gcp/packer.json.tmpl
```diff
@@ -13,10 +13,13 @@
...
     "network": "",
     "project_id": "",
-    "zone": ""
+    "zone": "",
+    "disk_size": "80",
+    "disk_type": "pd-standard"
   },
   "builders": [
     {
+       "name": "{{(user `distribution`) | lower}}-{{user `distribution_version`}}{{user `build_name_extra`}}",
```
alejandroEsc, line 22: is name a valid field here? if so how do I find out these "default" fields I am looking at this btw: `https://www.packer.io/plugins/builders/googlecompute`

faiq (approved):

pkg/packer/manifests/gcp/packer.json.tmpl
supershal: I agree, its hard to find documentation for `name`. some references: https://www.packer.io/docs/templates/hcl_templates/contextual-variables#build-variables
and
https://www.packer.io/docs/templates/legacy_json_templates/builders#named-builds

alejandroEsc (approved):
Lgtm

PR #400, opened 2022-06-30
chore: sign apply binaries

alejandroEsc:
Allows having to sign darwin images.

* https://jira.d2iq.com/browse/D2IQ-88902

```release-note

```

### DIFF
--- a/.github/workflows/sign-darwin.yml
+++ b/.github/workflows/sign-darwin.yml
@@ -1,0 +1,61 @@
+---
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'release tag to sign'
+        required: false
+        type: string
+
+jobs:
+  sign-binary:
+    runs-on: macos-latest
+    steps:
+      - uses: dsaltares/fetch-gh-release-asset@1.0.0
+        with:
+          version: "tags/${{ inputs.tag }}"
+          file: "konvoy-image-bundle-${{ inputs.tag }}_darwin_amd64.tar.gz"
+
+      - name: Extract the pulled file
+        run: tar -xvzf "konvoy-image-bundle-${{ inputs.tag }}_darwin_amd64.tar.gz" -C "konvoy-image-bundle_darwin_amd64"
+
+      - name: Remove original downloaded file
+        run: rm "konvoy-image-bundle-${{ inputs.tag }}_darwin_amd64.tar.gz"
+
+      - name: Fetch current main directory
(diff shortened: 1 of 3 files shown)


.github/workflows/sign-darwin.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          brew tap mitchellh/gon
+          brew install mitchellh/gon/gon
+
+      - name: Run Gon to notarize
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          AC_USERNAME: ${{ secrets.AC_USERNAME }}
+          AC_PASSWORD: ${{ secrets.AC_PASSWORD }}
+        run: gon ./gon.hcl -log-level=info -log-json ./config.hcl
```
faiq, line 49: where is config.hcl coming from?

.github/workflows/sign-darwin.yml
```diff
@@ -0,0 +1,61 @@
+---
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'release tag to sign'
+        required: false
+        type: string
+
+jobs:
```
faiq, line 10: let's move this whole thing to `release-main.yaml`

.github/workflows/sign-darwin.yml
alejandroEsc: I believe this is a file created for the logs: https://github.com/mitchellh/gon/blob/master/README.md

.github/workflows/sign-darwin.yml
alejandroEsc: So this requires inputs to work, we can and maybe should call these on another workflow like that found in the release-main process. At least that was my hope. Here at least this allows us a repeatable mechanism to get this to work.

faiq: The inputs can be populated by the release-please task outputs the `bump-kib` job does this. 

Keeping this here while it's draft is fine, but I really want to see it in the other file as another workflow.

alejandroEsc: Sure thing. There were some difficulties in the release job process itself that kept me from moving forward to that approach, it has been a while but I tried to communicate my frustration with that approach.

supershal:
Closing the ticket as per discussion. The notarizing mac binary in github public repo is not secure enough as it may expose apple credentials. Once we finalize the approach, we can revisit this PR

PR #406, opened 2022-07-06
fix: install cloud-init packages for vsphere in offline mode

supershal:
- install cloud-init packages for vsphere in offline mode

The vsphere airgapped test appear to be false positive. This package was pre-installed in the base-os template so the ansible playbooks always skipped them without searching for them in the `offline` repo.

Currently we are still have not added such fix for other OS than Redhat and centos, GPU etc. 
As we add more code to support offline image building in future, we can unintentionally setting `offline` repo. We will need to think about globally enable offline repo for airgapped mode.

### DIFF
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -29,6 +29,8 @@
   yum:
     name: "{{ packages }}"
     state: present
+    enablerepo: "{{ 'offline' if offline_mode_enabled else '' }}"
+    disablerepo: "{{ '*' if offline_mode_enabled else '' }}"
   vars:
     packages:
       - cloud-init


faiq (approved):

dkoshkin (approved):

ansible/roles/providers/tasks/vmware.yml
```diff
@@ -29,6 +29,8 @@
   yum:
     name: "{{ packages }}"
     state: present
+    enablerepo: "{{ 'offline' if offline_mode_enabled else '' }}"
+    disablerepo: "{{ '*' if offline_mode_enabled else '' }}"
   vars:
     packages:
       - cloud-init
```
dkoshkin, line 36: I also see that these packages are already included in our air-gapped artifacts https://github.com/mesosphere/dkp-release/blob/1d49a12fb7147e7b4863f14043e9806ee279bd95/os-packages-bundles/Makefile#L47

supershal: Thanks for checking.

PR #412, opened 2022-07-15
fix: --instance-type and other AWS flags not being set

dkoshkin:
Removed the problematic code that reinitialized the flags. This was broken for a long time, but it was exposed by https://github.com/mesosphere/konvoy-image-builder/commit/e0dcc561e0286db15ac480f55e8a3291e6e5f544 when we switched away from using `--aws-instance-type`.

Also added a new unit test for RHEL 8.4 NVIDIA.

* https://jira.d2iq.com/browse/D2IQ-91591

```release-note

```


dkoshkin:
Need additional changes, will reopen

PR #418, opened 2022-07-19
fix: gon release fixes

faiq:
adding a zip should notarize the file now also avoids a mistake in the release name.

```release-note

```

### DIFF
--- a/.github/workflows/release-main.yml
+++ b/.github/workflows/release-main.yml
@@ -150,7 +150,11 @@ jobs:
 
       - if: ${{ needs.release-please.outputs.release_created }}
         name: Tar signed bundle
-        run: tar czf konvoy-image-bundle-${{ needs.release-please.outputs.release_tag_name }}_darwin_amd64.tar.gz konvoy-image-bundle_darwin_amd64
+        run: |
+          unzip -o konvoy-image.zip
+          mv konvoy-image konvoy-image-bundle_darwin_amd64
+          mv konvoy-image-bundle_darwin_amd64 konvoy-image-bundle-${{ needs.release-please.outputs.release_tag_name }}_darwin_amd64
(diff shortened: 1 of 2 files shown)


gon.hcl
```diff
@@ -1,9 +1,15 @@
...
   password = "@env:AC_PASSWORD"
 }
+
 sign {
   application_identity = "Developer ID Application: Mesosphere Inc. (JQJDUUPXFN)"
 }
+
+zip {
+  "output_path": "konvoy-image.zip"
+}
```
dkoshkin, line 15: Why do we need this zip?

faiq: It is needed because it is what apple requires of things that can be notarized 

from gon docs 

> notarize (optional) - Settings for notarizing already built files. This is an alternative to using the source option. This option can be repeated to notarize multiple files.

>    path (string) - The path to the file to notarize. This must be one of Apple's supported file types for notarization: dmg, pkg, app, or zip.

faiq: > zip (optional) - Settings related to creating a zip archive as output. A zip archive will only be created if this is specified. Note that zip archives don't support stapling, meaning that files within the notarized zip archive will require an internet connection to verify on first use.

this wont work for the airgap mac user though

.github/workflows/release-main.yml
```diff
@@ -150,7 +150,11 @@ jobs:
 
       - if: ${{ needs.release-please.outputs.release_created }}
         name: Tar signed bundle
-        run: tar czf konvoy-image-bundle-${{ needs.release-please.outputs.release_tag_name }}_darwin_amd64.tar.gz konvoy-image-bundle_darwin_amd64
+        run: |
+          unzip -o konvoy-image.zip
+          mv konvoy-image konvoy-image-bundle_darwin_amd64
```
dkoshkin, line 155: Does this need to be like:
```suggestion
          mv konvoy-image konvoy-image-bundle_darwin_amd64/konvoy-image
```

dkoshkin (approved):

.github/workflows/release-main.yml
faiq: mv should replace whatever is there

PR #422, opened 2022-07-20
Faiq/smart bumps

faiq:
**What problem does this PR solve?**:
Sets the bump PR correctly with use of a new makefile target defined in https://github.com/mesosphere/cluster-api-provider-preprovisioned/pull/129 using a new script.

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-NUMBER

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #423, opened 2022-07-20
fix: always use fix for CAPPP bumps

faiq:
For now, let's hard code to fix until we get #422 in a better state. 

```release-note

```


dkoshkin:
In the future, we may want to take the changes from https://github.com/mesosphere/konvoy-image-builder/pull/422 and move the individual steps from here to a script file https://github.com/mesosphere/konvoy-image-builder/blob/55652951ccdeb5d2deed886c2bcce22cdb9d5132/.github/workflows/release-main.yml#L54

dkoshkin (approved):

PR #427, opened 2022-07-20
fix: fix ref in release tag ci

faiq:
This fix should allow us to run the jobs on tags properly now 

```release-note

```

### DIFF
--- a/.github/workflows/release-tag.yaml
+++ b/.github/workflows/release-tag.yaml
@@ -1,7 +1,7 @@
 on:
   push:
-    tags:        
-      - '*' 
+    tags:
+      - '*'
 
 name: upload-release-artifacts
 jobs:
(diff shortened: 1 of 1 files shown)


dkoshkin:
These changes look good but I don't think its enough, I would have still expected it to get triggered.
Maybe we need https://github.com/actions/runner/issues/1007#issuecomment-808904408?

This is Sus
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore
>If you define only tags/tags-ignore or only branches/branches-ignore, the workflow won't run for events affecting the undefined Git ref.

faiq:
I think the issue isn't going to apply to us. It looks like the original poster of the issue was using `create` events. The documentation seems like it supports the `tags` filter on `push` events.

dkoshkin:
Lets try it.

PR #433, opened 2022-07-20
fix: gon.hcl for binary signing

faiq:
makes gon.hcl work for binary signing 

```release-note

```

### DIFF
--- a/gon.hcl
+++ b/gon.hcl
@@ -11,5 +11,5 @@ sign {
 }
 
 zip {
-  "output_path": "konvoy-image.zip"
+  output_path = "konvoy-image.zip"
 }


supershal (approved):

Makefile
```diff
@@ -352,7 +353,7 @@ build.snapshot:
 	#                that it uses as its docker build context ;(.
 	mkdir -p bin
 	cp dist/konvoy-image_linux_amd64_v1/konvoy-image bin/konvoy-image
-	goreleaser --parallelism=1 --skip-publish --snapshot --rm-dist
+	goreleaser --parallelism=$(GORELEASER_PARALLELISM) --skip-publish --snapshot --rm-dist
```
dkoshkin, line 356: Since the feedback cycle is pretty long can we revert this change and only do the `gon` change.
Building these takes lots of resources and this may introduce additional errors in CI.

faiq: :cry:

faiq: it might be worth taking the flag out entirely, the default is 

```
  -p, --parallelism int              Amount tasks to run concurrently (default: number of CPUs)
  ```

faiq: its ok ill remove

dkoshkin (approved):

PR #449, opened 2022-08-01
fix: use new rhel7 template

faiq:
Uses the correct disk for the rhel 7 template. The template is now using this disk instead based on some configuration changes we made.

```release-note

```

### DIFF
--- a/images/ova/rhel-79.yaml
+++ b/images/ova/rhel-79.yaml
@@ -13,7 +13,6 @@ packer:
   insecure_connection: "false"
   network: ""
   resource_pool: ""
-  template: "base-rhel-7"
   vsphere_guest_os_type: "rhel7_64Guest"
   guest_os_type: "rhel7-64"
   # goss params
--- a/images/ova/rhel-79.yaml
+++ b/images/ova/rhel-79.yaml
@@ -13,6 +13,7 @@ packer:
   insecure_connection: "false"
   network: ""
   resource_pool: ""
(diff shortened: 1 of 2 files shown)


alejandroEsc (approved):
lgtm

images/ova/rhel-79.yaml
```diff
@@ -13,7 +13,7 @@ packer:
   insecure_connection: "false"
   network: ""
   resource_pool: ""
-  template: "base-rhel-7"
+  template: "base-rhel-7-sortega"
```
dkoshkin, line 16: What is "sortega"? This is customer facing file, is this d2iq specific?

dkoshkin (changes requested):

images/ova/rhel-79.yaml
faiq: This is the name of our base image. Our ones named `base-rhel-7` were not configured correctly.

dkoshkin: Thats what I figured, but is "sortega" meaningful outside of D2IQ and our template? This is changing an external file used by our customers. I think it would be unexpected for them that an argument changed.

supershal: Thanks. this is a good point. I can rename the `base-rhel-7-sortega` to `base-rhel-7` in vsphere.

supershal:
closing this PR. I will rename the template in the vsphere instead.

PR #452, opened 2022-08-02
fix: add overrides, work-dir, and extra-args flags to upload artifacts

faiq:
Allows a user to pass overrides to the upload command. This is necessary for uploading fips bundles for containerd. It also brings the experience closer to the other commands. 
Finally, we have some refactors for common functionality to create the `extraVars` file that build and artifacts command share.

* https://jira.d2iq.com/browse/D2IQ-92024

```release-note

```


cmd/konvoy-image/cmd/args.go
```diff
@@ -16,3 +16,7 @@ func addClusterArgs(fs *pflag.FlagSet, kubernetesVersion, containerdVersion *str
 	fs.StringVar(kubernetesVersion, "kubernetes-version", "", "The version of kubernetes to install. Example: 1.21.6")
 	fs.StringVar(containerdVersion, "containerd-version", "", "the version of containerd to install")
 }
+
+func addWorkDirArg(fs *pflag.FlagSet, workDir *string) {
+	fs.StringVar(workDir, "work-dir", "", "path to custom work directory generated by the command")
+}
```
dkoshkin, line 22: Im not understanding what this flag is for, somehow related to the `generate` command?
```
➜  konvoy-image-bundle-v1.19.7_darwin_amd64 ./konvoy-image generate aws --help
generate files relating to building aws images

Usage:
  konvoy-image generate aws <image.yaml> [flags]

Examples:
aws --region us-west-2 --source-ami=ami-12345abcdef images/ami/centos-79.yaml

Flags:
      --ami-groups stringArray           a list of AWS groups which are allowed use the image, using 'all' result in a public image
      --ami-regions stringArray          a list of regions to publish amis
      --ami-users stringArray            a list AWS user accounts which are allowed use the image
      --containerd-version string        the version of containerd to install
      --extra-vars strings               flag passed Ansible's extra-vars
  -h, --help                             help for aws
      --instance-type string             instance type used to build the AMI; the type must be present in the region in which the AMI is built
      --kubernetes-version string        The version of kubernetes to install. Example: 1.21.6
      --overrides strings                a comma separated list of override YAML files
      --region string                    the region in which to build the AMI
      --source-ami string                the ID of the AMI to use as the source; must be present in the region in which the AMI is built
      --source-ami-filter-name string    restricts the set of source AMIs to ones whose Name matches filter
      --source-ami-filter-owner string   restricts the source AMI to ones with this owner ID

Global Flags:
      --color     enable color output (default true)
  -v, --v int     select verbosity level, should be between 0 and 6
      --verbose   enable debug level logging (same as --v 5)
```

dkoshkin: I see it also in `build`
```
➜  konvoy-image-bundle-v1.19.7_darwin_amd64 ./konvoy-image build aws images/ami/centos-79.yaml --work-dir dkoshkin
INFO[0000] using workDir provided by --work-dir flag: dkoshkin
ERRO error during run: open dkoshkin/ansible_vars.yaml: no such file or directory
error encountered: exit status 3
```

faiq: kib creates a directory where it puts all of the files it needs to run ansible and packer commands in `work-dir`. Here's how it is used in [build](https://github.com/mesosphere/konvoy-image-builder/blob/e44cceb9b97b95778fdfb58f96110405f0ce2883/pkg/app/build.go#L103)


 We're using it here because I'm using a file for extra-vars. This file allows us to handle nested maps a lot easier.

faiq: normally this is `$(pwd)/work/<build-name>`

pkg/app/artifacts.go
```diff
@@ -15,18 +17,40 @@ type ArtifactsCmdFlags struct {
...
-	playbookOptions, err := playbookOptionsFromFlag(artifactFlags)
+type ArtifactUploader struct {
+	workDir string
+}
+
+func NewArtifactUploader(buildName string) (*ArtifactUploader, error) {
+	if buildName == "" {
+		name := "artifact-upload"
+		buildName = fmt.Sprintf("%s-%d", name, time.Now().Unix())
+	}
```
supershal, line 33: nit: moving this logic to the caller could read better.

pkg/app/artifacts.go
```diff
@@ -43,12 +67,32 @@ func playbookOptionsFromFlag(artifactFlags ArtifactsCmdFlags) (*ansible.Playbook
...
 		return nil, fmt.Errorf("failed to find absolute path for --container-images-dir %w", err)
 	}
+	passedUserArgs := map[string]interface{}{
+		"os_packages_local_bundle_file":  osPackagesBundleFile,
+		"containerd_local_bundle_file":   containerdBundleFile,
+		"pip_packages_local_bundle_file": pipPackagesBundleFile,
+		"images_local_bundle_dir":        containerImagesDir,
+	}
+
+	if err = mergeUserOverridesToMap(artifactFlags.Overrides, passedUserArgs); err != nil {
```
supershal, line 77: since the `passedUserArgs` flags are marked as `required`, should they get overriden from the override file and extra-vars? 
We should either not mark them as `required` or give them highest precedence if they are `required` . wdyt?

faiq: Agreed. This code block should be the last one and given highest precedence

dkoshkin:
What do you think about renaming the PR to "fix: add overrides, work-dir, and extra-args flags to upload artifacts](https://github.com/mesosphere/konvoy-image-builder/pull/452#top)", this is really fixing a bug by adding some new flags.
It would allow us to release a patch version that can then be used along with DKP v2.3.0?

faiq:
Changed the name. I can give the code a final walk through before merging if you have any questions

pkg/app/build_test.go
```diff
@@ -52,4 +52,59 @@ func TestMapMerge(t *testing.T) {
 	if assert.Contains(t, m1, "four") {
 		assert.Equal(t, m1["four"], m3["four"])
 	}
+
+}
+
+func TestMapMergeForArtifacts(t *testing.T) {
```
supershal, line 58: Thanks for the unit test.

supershal (approved):

faiq:
Waiting for @dkoshkin approval before merging

dkoshkin (approved):

dkoshkin:
Great work, as long as we document that this version should be used for DKP `v2.3.0` I don't think we need the workaround anymore docs.

PR #450, opened 2022-08-02
Ubuntu fixes for vSphere

markround:
**What problem does this PR solve?**:

Fixes a couple of minor issues that prevented Ubuntu 20.04 images being built on vSphere:

- Ansible failed when stopping a service that did not exist on a stock install (hv-kvp-daemon)
- /run is mounted with "noexec" by default in Ubuntu which prevented cloud-init scripts from running

**Which issue(s) does this PR fix?**:

No JIRA issue created (I can add one if required)

**Special notes for your reviewer**:

As discussed in #konvoy channel in D2IQ internal Slack

**Does this PR introduce a user-facing change?**:

No, although a new ubuntu image definition could be added to "images/ova" which would make this visible. I have not done so as part of this PR though.

PR #458, opened 2022-08-04
fix: ubuntu  for vsphere 

markround:

With these three patches (and the ones previously merged) it is now possible to build Ubuntu 20.04 OVAs for vSphere. I have tested these images by building DKP clusters with them, both management and workload and have deployed a full Kommander installation on top.

No JIRA issue for these patches

You may find the following image YAML useful if you want to verify the images yourself:

```yaml
---
download_images: true
build_name: "ubuntu-2004"
packer_builder_type: "vsphere" 
packer:
  cluster: ""
  datacenter: ""
  datastore: ""
  folder: ""
  insecure_connection: "false"
  network: ""
  resource_pool: ""
  template: "base-ubuntu-2004"
  vsphere_guest_os_type: "ubuntu64Guest"
  guest_os_type: "ubuntu-64"
# goss params
  distribution: "Ubuntu"
  distribution_version: "20.04"
```

You may also want to set the following override if testing with DKP 2.2.2:

```yaml
kubernetes_version: 1.22.8
```

Please feel free to reach out to me on D2IQ slack (@mdr) if you have any questions.

No.

### DIFF
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -82,6 +82,16 @@
     group: root
     mode: 0644
 
+- name: Remove subiquity-disable-cloudinit-networking.cfg
+  file:
+    path: /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg
+    state: absent
+
+- name: Remove 99-installer.cfg
+  file:
+    path: /etc/cloud/cloud.cfg.d/99-installer.cfg
+    state: absent
+
 # this program used by ds-identify to determine whether or not the
 # VMwareGuestInfo datasource is useable.
 - name: Create ds-check program to verify VMwareGuestInfo datasource
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -142,3 +142,13 @@
     owner: root
     group: root
     mode: 0644
+
+- name: Get a list of services
+  service_facts:
+
+- name: Disable UFW firewall on Ubuntu/Debian
+  systemd:
+    name: ufw
+    state: stopped
+    enabled: false
+  when: ('ufw.service' in ansible_facts.services) and (ansible_os_family == "Debian")
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -152,3 +152,21 @@
     state: stopped
     enabled: false
   when: ('ufw.service' in ansible_facts.services) and (ansible_os_family == "Debian")
+
+# See https://kb.vmware.com/s/article/82229
+# From systemd docs: 
+# "If a valid D-Bus machine ID is already configured for the system, 
+# the D-Bus machine ID is copied and used to initialize the machine ID in /etc/machine-id"
+# This needs to be reset/truncated as well on Ubuntu.
+- name: Truncate D-Bus machine-id
+  file:
+    path: /var/lib/dbus/machine-id
+    state: absent
+  when: ansible_os_family == "Debian" 
+
+- name: Link D-Bus machine-id to /etc/machine-id
+  file:
+    path: /var/lib/dbus/machine-id
+    state: link
+    src: /etc/machine-id
+  when: ansible_os_family == "Debian"


ansible/roles/networking/tasks/main.yaml
```diff
@@ -57,6 +57,16 @@
...
 
+- name: Get a list of services
+  service_facts:
+
+- name: Disable UFW firewall on Ubuntu/Debian
+  systemd:
+    name: ufw
+    state: stopped
+    enabled: false
+  when: ('ufw.service' in ansible_facts.services) and (ansible_os_family == "Debian")
```
faiq, line 68: i think we want to add packer.build_type = 'vsphere' or something like that too

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -42,6 +42,19 @@
       state: touch
       mode: "{{ machine_id_mode }}"
 
+- name: Truncate D-Bus machine-id
```
dlipovetsky, line 45: Please add a comment explaining why we need to do this.

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -42,6 +42,19 @@
...
+  file:
+    path: /var/lib/dbus/machine-id
+    state: absent
+  when: ansible_os_family == "Debian"
+
+- name: Link D-Bus machine-id to /etc/machine-id
```
dlipovetsky, line 51: Please add a comment explaining why we need to do this.

ansible/roles/providers/tasks/vmware.yml
```diff
@@ -82,6 +82,24 @@
     group: root
     mode: 0644
 
+- name: Ensure VMware customisations are enabled
+  copy:
+    src: files/etc/cloud/cloud.cfg.d/10_vmware.cfg
+    dest: /etc/cloud/cloud.cfg.d/10_vmware.cfg
+    owner: root
+    group: root
+    mode: 0644
```
faiq, line 91: What are the VMware customizations? Can you link to some documentation to make sure we want this across all OSes for vsphere?

ansible/roles/networking/tasks/main.yaml
markround: I can do that, but we don't currently check for that when disabling e.g. iptables service on other distros?

ansible/roles/sysprep/tasks/main.yml
markround: Sure. It's in the commit message but I'll drop something in the file itself.

ansible/roles/networking/tasks/main.yaml
dlipovetsky: Good point, @faiq .

I'd amend that: if this task is specific to Ubuntu _on VMWare_, I think it belongs in https://github.com/mesosphere/konvoy-image-builder/blob/e6fff29f0b373ad87684dbb5ef0316e9947d00df/ansible/roles/providers/tasks/vmware.yml, which is where the changes from #450 went.

ansible/roles/sysprep/tasks/main.yml
dlipovetsky: Is this only needed for Ubuntu on VMWare?

ansible/roles/sysprep/tasks/main.yml
dlipovetsky: Is this only needed for Ubuntu on VMWare?

ansible/roles/providers/tasks/vmware.yml
dlipovetsky: +1

ansible/roles/sysprep/tasks/main.yml
markround: It appears to be the case. My VMware RHEL builds work fine without it, I suspect it's purely an Ubuntu thing. I've only found references to it being a problem on Ubuntu (see the linked vmware KB article in the commit message for an example).

Oddly, I have one older physical Ubuntu 20.04 server at home where /var/lib/dbus/machine-id is already a symlink, perhaps it changed in a point release sometime?

ansible/roles/providers/tasks/vmware.yml
markround: Oh drat. I meant to only do this for Ubuntu. Good catch, let me fix this up and add some documentation links.

markround: Turns out it isn't needed for the Ubuntu build - I was basing it off https://kb.vmware.com/s/article/59557 but according to https://williamlam.com/2022/06/using-the-new-vsphere-guest-os-customization-with-cloud-init-in-vsphere-7-0-update-3.html it is only used to switch to a new customization option available in vSphere >= 7.0u3. Leaving it out does not affect anything on Ubuntu 20.04 on my vSphere 7.0u3 environment, images still build and can be used to create clusters so I have removed it from the patch set. 

It might be something to consider in the future though, if the "traditional GOSC" customization method becomes deprecated.

ansible/roles/sysprep/tasks/main.yml
markround: I have updated the commit message with another link to https://www.freedesktop.org/software/systemd/man/systemd-machine-id-setup.html which explains how /etc/machine-id can be recreated from the DBUS machine-id if it is not also cleaned. Also added a comment above the relevant blocks in Ansible. 

It does not affect RHEL as on my RHEL base images, /var/lib/dbus/machine-id is also already a symlink. But this is not the case of a stock Ubuntu 20.04 install from the latest ubuntu-20.04.4-live-server-amd64.iso.

ansible/roles/sysprep/tasks/main.yml
markround: See above comment. I believe this may be required for _all_ Ubuntu images to ensure consistency. 

In some Ubuntu 20.04 systems in my homelab, this symlink seems to be already present but the common denominator is that all these were created from the "desktop" ISO. Reading a comment on an old Launchpad bug (https://bugs.launchpad.net/ubuntu/+source/dbus/+bug/1508697/comments/4) would explain this:

>  * /usr/lib/tmpfiles.d/dbus.conf will create the /var/lib/dbus/machine-id symlink to /etc/machine-id. I. e. this would happen on "first boot" *if* /var/lib/dbus/machine-id does not exist yet. This is what happens if you install from a desktop CD: livecd-rootfs removes machine-id from the image, so the first boot goes through that tmpfiles.d snippet and creates the symlink.
> 
>  * If you install server, that actually apt-get installs stuff including dbus. dbus.postinst calls "dbus-uuidgen --ensure" which according to NEWS
> 
>      • in "dbus-uuidgen --ensure", try to copy systemd's /etc/machine-id
>        to /var/lib/dbus/machine-id instead of generating an entirely new ID
>        (fd.o #77941, Simon McVittie)

Which explains why on a stock server install, the symlink is not present and it is a regular file with the same content as /etc/machine-id (https://www.freedesktop.org/software/systemd/man/systemd-machine-id-setup.html explains how /etc/machine-id can be recreated from the DBUS machine-id if it is not also cleaned.)

My thinking on this is that as RHEL seems to do the symlink anyway, and some Ubuntu installs also will do this depending on the image used, it's good to ensure it's done in a consistent way across all variations of Ubuntu images.

ansible/roles/networking/tasks/main.yaml
markround: I don't think disabling UFW is specific to just VMware. It should be disabled for the same reason we also disable iptables/other firewall services because it can cause conflicts and node communication issues. But I'm happy to move it to the provider task or add a `packer.build_type = 'vsphere'` if you think it's best to localize the changes.

alejandroEsc: I believe since the title of this PR is fixes for vsphere the expectation would be that these would be limited to vsphere. However its cool if this is a bug for other cloud-providers, is this what you are suggesting @markround ?

ansible/roles/sysprep/tasks/main.yml
alejandroEsc: I think its fair to run this only on the vmware case. can we add a check for this as @dlipovetsky suggests?

ansible/roles/networking/tasks/main.yaml
faiq: please localize the changes.

ansible/roles/sysprep/tasks/main.yml
faiq: i think moving to vmware provider makes sense, with a note.

ansible/roles/networking/tasks/main.yaml
markround: Change moved to vmware provider as discussed. Thanks for the feedback!

ansible/roles/sysprep/tasks/main.yml
markround: Change moved to vmware provider as discussed. Thanks for the feedback!

faiq (approved):

supershal (approved):

PR #459, opened 2022-08-07
fix(azure): support dry run in image creation

jkoelker:

Adds `DryRun` support to azure now that https://github.com/hashicorp/packer-plugin-azure/pull/186 is merged.

* https://jira.d2iq.com/browse/D2IQ-85259

```release-note

```


jkoelker:
Alas, it looks like while it skips image creation, packer still panics, fixing upstream, I ran this build from a plugin release in my fork, which is why it passed in the last run ;)

supershal (approved):
🎉 . Thank you for the right upstream fix.

faiq (approved):

PR #469, opened 2022-08-19
fix: remove builder user in sysprep

faiq:
https://jira.d2iq.com/browse/D2IQ-92459

```release-note

```

### DIFF
--- a/ansible/roles/sysprep/tasks/main.yml
+++ b/ansible/roles/sysprep/tasks/main.yml
@@ -157,6 +157,13 @@
     - path: /root/.ssh/authorized_keys
   when: ansible_os_family == "Flatcar"
 
+- name: Remove Builder User
+  user:
+    name: builder
+    state: absent
+    remove: yes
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):

supershal (approved):
Thanks for fixing this.

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -157,6 +157,13 @@
     - path: /root/.ssh/authorized_keys
   when: ansible_os_family == "Flatcar"
 
+- name: Remove Builder User
+  user:
+    name: builder
```
supershal, line 162: Just wanted to note that a customer's base template might not use the same bast template as ours and will not have `builder` user.
- The user module takes care if the `builder` user does not exists and still runs this task successfully. https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html#parameter-state
- It would be better if the list of users can be taken as array variable and execute this task for all the packer builder type (vmware or aws or azure). we can add `builder` as default user in the variable. 
-

PR #474, opened 2022-08-24
feat: disable password ssh auth

faiq:
Reverts mesosphere/konvoy-image-builder#469

### DIFF
--- a/ansible/roles/sysprep/tasks/main.yml
+++ b/ansible/roles/sysprep/tasks/main.yml
@@ -196,3 +196,4 @@
     line: "PasswordAuthentication no"
     state: present
     backup: yes
+  when: ansible_os_family != "Flatcar"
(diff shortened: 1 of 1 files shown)


faiq:
this made tests fail. I should have caught that 

refactor like this.
https://stackoverflow.com/a/58257737/3542352

supershal (approved):

supershal (approved):

ansible/roles/sysprep/tasks/main.yml
```diff
@@ -195,3 +188,12 @@
     name: kubelet
     enabled: true
     state: started
+
+- name: Disable Password Authentication
```
dkoshkin, line 192: I _think_ this is ok since its in the `sysprep` role. 
Do we want to be more defensive and add `when: packer_builder_type.startswith('vsphere') or packer_builder_type.startswith('vmware')`?

dkoshkin (approved):

ansible/roles/sysprep/tasks/main.yml
faiq: Also thought about this.

I think this is a best practices sort of thing to disable password auth. If I understand correctly, CAPI providers leverage cloud-init to configure hosts. I think CAPPP uses ssh-key authentication

PR #476, opened 2022-08-25
feat: import container images with mindthegap

faiq:
uses mindthegap for k8s versions > 1.24 for import 

```release-note

```

### DIFF
--- a/Dockerfile.devkit
+++ b/Dockerfile.devkit
@@ -68,6 +68,7 @@ ARG USER_ID=0
 ARG GROUP_NAME=root
 ARG GROUP_ID=0
 ARG DOCKER_GID=0
+ARG MINDTHEGAP_VERSION=0.19.0
 
 COPY requirements.txt /tmp/
 COPY requirements-devkit.txt /tmp/
@@ -97,6 +98,7 @@ RUN apk add --no-cache \
     && rm -rf \
         /root/.cache
 
+RUN  curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz |tar xzf - -C /usr/local/bin
 RUN  curl -fsSL --output /tmp/goss.install  https://goss.rocks/install \
     && GOSS_VER=v0.3.16 GOSS_DST=/usr/local/bin bash /tmp/goss.install \
     && rm /tmp/goss.install
--- a/Dockerfile.devkit
+++ b/Dockerfile.devkit
@@ -98,6 +98,7 @@ RUN apk add --no-cache \
     && rm -rf \
         /root/.cache
 
+# hadolint ignore=DL4006
 RUN  curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz |tar xzf - -C /usr/local/bin
 RUN  curl -fsSL --output /tmp/goss.install  https://goss.rocks/install \
     && GOSS_VER=v0.3.16 GOSS_DST=/usr/local/bin bash /tmp/goss.install \
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -13,28 +13,16 @@
     - name: import images from files in the local images cache # noqa no-changed-when
       shell:
         executable: /bin/bash
-        cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
-      when:
-        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
-        - item.split('/')[-1].split('.')[1] |int < 24
+        cmd: |
+          ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
+          if [[ $? -ne 0 ]]; then
+            {{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }};
+          fi
       register: ctr_images_import
       failed_when: ctr_images_import.rc != 0
       changed_when: ctr_images_import.rc == 0
       with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
 
-    - name: import images from files to local images cache with mindthegap
-      shell:
-        executable: /bin/bash
-        cmd: mindthegap import image-bundle --image-bundle={{ item }}
-      when:
-        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
-        - item.split('/')[-1].split('.')[1] |int >= 24
-      register: mindthegap_images_import
-      failed_when: mindthegap_images_import.rc != 0
-      changed_when: mindthegap_images_import.rc == 0
-      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
-
     - name: remove mindthegap if present
-      command: rm /usr/local/bin/mindthegap || true
-
+      command: rm {{ mindthegap_binary_location_on_remote }} || true
   when: images_cache_find_result.files | length > 0
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -25,7 +25,7 @@
     - name: import images from files to local images cache with mindthegap
       shell:
         executable: /bin/bash
-        cmd: mindthegap import image-bundle {{ item }}
+        cmd: ${HOME}/bin/mindthegap import image-bundle {{ item }}
       when:
         - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
         - item.split('/')[-1].split('.')[1] |int >= 24
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -25,6 +25,6 @@
 
     - name: remove mindthegap if present
       file:
-        path: {{ mindthegap_binary_location_on_remote }}
+        path: "{{ mindthegap_binary_location_on_remote }}"
         state: absent
   when: images_cache_find_result.files | length > 0
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -24,5 +24,7 @@
       with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
 
     - name: remove mindthegap if present
-      command: rm {{ mindthegap_binary_location_on_remote }} || true
+      shell:
+        executable: /bin/bash
+        cmd: rm {{ mindthegap_binary_location_on_remote }} || true
   when: images_cache_find_result.files | length > 0
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -25,7 +25,7 @@
     - name: import images from files to local images cache with mindthegap
       shell:
         executable: /bin/bash
-        cmd: ${HOME}/bin/mindthegap import image-bundle {{ item }}
+        cmd: /tmp/bin/mindthegap import image-bundle {{ item }}
       when:
         - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
         - item.split('/')[-1].split('.')[1] |int >= 24
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -25,7 +25,7 @@
     - name: import images from files to local images cache with mindthegap
       shell:
         executable: /bin/bash
-        cmd: /tmp/bin/mindthegap import image-bundle --image-bundle={{ item }}
+        cmd: mindthegap import image-bundle --image-bundle={{ item }}
       when:
         - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
         - item.split('/')[-1].split('.')[1] |int >= 24
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -18,9 +18,6 @@
           if [[ $? -ne 0 ]]; then
             ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
           fi
-      register: ctr_images_import
-      failed_when: ctr_images_import.rc != 0
-      changed_when: ctr_images_import.rc == 0
       with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
 
     - name: remove mindthegap if present
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -14,9 +14,9 @@
       shell:
         executable: /bin/bash
         cmd: |
-          ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
+          {{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }};
           if [[ $? -ne 0 ]]; then
-            {{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }};
+            ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
           fi
       register: ctr_images_import
       failed_when: ctr_images_import.rc != 0
@@ -24,7 +24,7 @@
       with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
 
     - name: remove mindthegap if present
-      shell:
-        executable: /bin/bash
-        cmd: rm {{ mindthegap_binary_location_on_remote }} || true
+      file:
+        path: {{ mindthegap_binary_location_on_remote }}
+        state: absent
   when: images_cache_find_result.files | length > 0
--- a/ansible/roles/images/tasks/load.yaml
+++ b/ansible/roles/images/tasks/load.yaml
@@ -9,12 +9,28 @@
   register: images_cache_find_result
   changed_when: false
 
-- name: import images from files in the local images cache  # noqa no-changed-when
-  shell:
-    executable: /bin/bash
-    cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+- block:
+    - name: import images from files in the local images cache # noqa no-changed-when
+      shell:
+        executable: /bin/bash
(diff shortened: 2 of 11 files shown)


pkg/packer/manifests/azure/packer.json.tmpl
```diff
@@ -166,13 +166,26 @@
...
         "type": "file",
         "source": "/usr/local/bin/goss",
         "destination": "/tmp/.goss-dir/goss",
         "direction": "upload",
         "max_retries": 10
     },
+    {
+        "type": "file",
+        "source": "/usr/local/bin/mindthegap",
+        "destination": "${HOME}/bin/mindthegap",
```
faiq, line 185: I'm realizing this is going to fail on flatcar because of the ROFS

Dockerfile.devkit
```diff
@@ -68,6 +68,7 @@ ARG USER_ID=0
 ARG GROUP_NAME=root
 ARG GROUP_ID=0
 ARG DOCKER_GID=0
+ARG MINDTHEGAP_VERSION=0.19.0
```
dkoshkin, line 71: Whatever you decide, lets just keep it consistent please https://github.com/mesosphere/dkp-release/pull/25/files#r955354523
```suggestion
ARG MINDTHEGAP_VERSION=v0.19.0
```

Dockerfile.devkit
```diff
@@ -97,6 +98,7 @@ RUN apk add --no-cache \
     && rm -rf \
         /root/.cache
 
+RUN  curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz |tar xzf - -C /usr/local/bin
```
dkoshkin, line 101:
```suggestion
RUN  curl -Lf https://github.com/mesosphere/mindthegap/releases/download/"${MINDTHEGAP_VERSION}"/mindthegap_"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz |tar xzf - -C /usr/local/bin
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,28 @@
...
-    executable: /bin/bash
-    cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+- block:
+    - name: import images from files in the local images cache # noqa no-changed-when
+      shell:
+        executable: /bin/bash
+        cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+      when:
+        - item | regexp(^1\.(\d)+\.(\d)+_images\.tar\.gz$)
+        - item.split('.')[1] |int < 24
```
dkoshkin, line 19: @faiq @supershal @dlipovetsky  what do you think about taking this opportunity and renaming these k8s >=v1.24 image files?

Currently its:
```
➜  ~ aws s3 ls s3://downloads.mesosphere.io/dkp/airgapped/kubernetes-images/
2022-04-04 12:58:39  205642603 1.21.6_images.tar.gz
2022-04-04 13:00:22  208045884 1.21.6_images_fips.tar.gz
```

Maybe we can start creating any packages for >1.24 
```
kubernetes-images-1.21.6-d2iq.1.tar.gz
kubernetes-images-1.21.6-d2iq.1_fips.tar.gz
```

This `when` condition would be much simpler to write.
This would also aligns it better with:
```
➜  ~ aws s3 ls s3://packages.d2iq.com/dkp/containerd/
2022-05-25 16:46:22          0
2022-06-01 09:51:40   50027539 containerd-1.4.13-d2iq.1-centos-7.9-x86_64.tar.gz
2022-06-01 09:51:56   54357021 containerd-1.4.13-d2iq.1-centos-7.9-x86_64_fips.tar.gz
```

dkoshkin: This also has the benefit that it will not break any of our e2e automation. As we can still have both image tars for 1.24.4

Dockerfile.devkit
faiq: I don't understand isn't it already consistent with what is in dkp-release?

pkg/packer/manifests/aws/packer.json.tmpl
```diff
@@ -195,13 +195,26 @@
...
+      ]
+    },
     {
         "type": "file",
         "source": "/usr/local/bin/goss",
         "destination": "/tmp/.goss-dir/goss",
         "direction": "upload",
         "max_retries": 10
     },
+    {
```
supershal, line 211: what do you think to upload the `mindthdgap` binary in ansible? you can add retries and remove the binary once upload is finished. 
I was planning to try the same for `goss` binary.

ansible/roles/images/tasks/load.yaml
faiq: I like the idea of a name change

pkg/packer/manifests/aws/packer.json.tmpl
faiq: I don't get what the benefit would be of using ansible. Packer accomplishes the job for us

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,28 @@
...
-  shell:
-    executable: /bin/bash
-    cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+- block:
+    - name: import images from files in the local images cache # noqa no-changed-when
+      shell:
+        executable: /bin/bash
+        cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+      when:
+        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
```
supershal, line 18: There are multiple ways to support backward compatibility.  below are the alternatives i am thinking.
wdyt ?
```yaml
when: "{{ kubernetes_version is version('1.24.4, '>=', version_type='semver') }}"
```
https://github.com/mesosphere/konvoy-image-builder/blob/main/images/common.yaml#L2
https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#comparing-versions

OR 
- try both ways to upload images
- ignore errors
- add new task to fail if `ctr_images_import.rc !=0 && mindthegap.rc !=0`
I am trying above to see if its possible.

faiq:
all of our pre-provisioned machines need `mindthegap`. Maybe we should consider the ansible approach @supershal was talking about :thinking:

pkg/packer/manifests/aws/packer.json.tmpl
supershal: I am thinking 
- ansible playbook is common for any providers so code change stays at one place.
- the source binary dont have to be at the `/usr/local/bin` . however both packer and ansible can use variable to represent source path. .i.e. ansible playbook does not need to assume that binary exists at `/tmp/bin/mindthegap` 
- remove `mindthegap` binary after ansible task is finished as we dont want to leave it behind in the AMI.

faiq: We need to keep the `mindthegap` binary for pre-provisioned deployments. 

> the source binary dont have to be at the /usr/local/bin
this is where the mindthegap lives on the devkit container and production container. If we make an `arm` version we will still need to package both.

> ansible playbook is common for any providers so code change stays at one place.

this is the biggest selling point to me

ansible/roles/images/tasks/load.yaml
dlipovetsky: +1

And we can take this opportunity to change `_fips` to `-fips` :partying_face:

supershal: +1

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,28 @@
...
+      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
+
+    - name: import images from files to local images cache with mindthegap
+      shell:
+        executable: /bin/bash
+        cmd: mindthegap import image-bundle --image-bundle={{ item }}
+      when:
+        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
+        - item.split('/')[-1].split('.')[1] |int >= 24
+      register: ctr_images_import
```
supershal, line 32: should rename to `mindthegap_images_import`

supershal (approved):
nice improvements and backward compatibility.

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,32 @@
...
+      shell:
+        executable: /bin/bash
+        cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+      when:
+        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
+        - item.split('/')[-1].split('.')[1] |int < 24
+      register: ctr_images_import
+      failed_when: ctr_images_import.rc != 0
+      changed_when: ctr_images_import.rc == 0
+      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
```
dkoshkin, line 23: Users are already including additional bundles that they have built with `docker save` and may push mindthegap bundles either generated by DKP's release process or the dkp CLI, and these will not conform to any regex.

```suggestion
      when:
        mindthegap_images_import.rc != 0
      failed_when: False
      register: ctr_images_import
      changed_when: ctr_images_import.rc == 0
      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,32 @@
...
+      shell:
+        executable: /bin/bash
+        cmd: mindthegap import image-bundle --image-bundle={{ item }}
+      when:
+        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
+        - item.split('/')[-1].split('.')[1] |int >= 24
+      register: mindthegap_images_import
+      failed_when: mindthegap_images_import.rc != 0
+      changed_when: mindthegap_images_import.rc == 0
+      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
```
dkoshkin, line 35:
```suggestion
      failed_when: False
      register: mindthegap_images_import
      changed_when: mindthegap_images_import.rc == 0
      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,32 @@
...
+
+    - name: remove mindthegap if present
+      command: rm /usr/local/bin/mindthegap || true
```
dkoshkin, line 38: Why do we need to remove it? Keeping it on the host will make the runs a little faster since it won't need to recopy if there are no changes.

```suggestion
    - name: remove mindthegap if present
      file:
        path: "{{ mindthegap_binary_location_on_remote }}"
        state: absent
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,32 @@
...
+        - item.split('/')[-1].split('.')[1] |int < 24
+      register: ctr_images_import
+      failed_when: ctr_images_import.rc != 0
+      changed_when: ctr_images_import.rc == 0
+      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
+
+    - name: import images from files to local images cache with mindthegap
+      shell:
+        executable: /bin/bash
+        cmd: mindthegap import image-bundle --image-bundle={{ item }}
```
dkoshkin, line 28: Avoids errors when `/usr/local/bin/` is not in the `PATH`.

```suggestion
        cmd: "{{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }}"
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,32 @@
...
+      when:
+        - item.split('/')[-1] | regex_search('^1\.(\d)+\.(\d)+_images\.tar\.gz$')
+        - item.split('/')[-1].split('.')[1] |int >= 24
+      register: mindthegap_images_import
+      failed_when: mindthegap_images_import.rc != 0
+      changed_when: mindthegap_images_import.rc == 0
+      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
+
```
dkoshkin, line 36:
```suggestion
  - name: verify importing the images was successful 
    ansible.builtin.fail:
      msg: "Failed to load images into Containerd"
    when: 
      - mindthegap_images_import.rc != 0
      - ctr_images_import.rc != 0

```

ansible/roles/images/tasks/load.yaml
dkoshkin: I think this block should also move after `mindthegap` since going forward we prefer `mindthegap` and don't want to run this task and fail every time.

ansible/roles/images/tasks/load.yaml
supershal: IMHO, we should not leave leftover binaries (not only mindthegap) in the final image for reasons of security and minimize the size of the image. Some environments where network copy is not as good as S3 (like vsphere), it takes time to clone/copy AMI/image to each VM.

supershal (approved):

ansible/roles/offline/tasks/upload.yaml
```diff
@@ -71,6 +76,7 @@
         src: "{{ offline.os_packages_remote_bundle_path }}/offline-repo.tar.gz"
         dest: "{{ offline.os_packages_remote_filesystem_repo_path }}"
         remote_src: yes
+        mode: 0755
```
dkoshkin, line 79: Why is this needed?

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,22 @@
...
+          fi
+      register: ctr_images_import
+      failed_when: ctr_images_import.rc != 0
+      changed_when: ctr_images_import.rc == 0
+      with_items: "{{  images_cache_find_result.files | map(attribute='path') | list }}"
+
+    - name: remove mindthegap if present
+      shell:
+        executable: /bin/bash
+        cmd: rm {{ mindthegap_binary_location_on_remote }} || true
```
dkoshkin, line 29: Should we also move this out of the `block` and always delete it?
```suggestion
    - name: remove mindthegap if present
      file:
        path: "{{ mindthegap_binary_location_on_remote }}"
        state: absent
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,22 @@
...
+      shell:
+        executable: /bin/bash
+        cmd: |
+          ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
+          if [[ $? -ne 0 ]]; then
+            {{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }};
+          fi
+      register: ctr_images_import
+      failed_when: ctr_images_import.rc != 0
+      changed_when: ctr_images_import.rc == 0
```
dkoshkin, line 23: I don't think we need these and Ansible will take care of it.
```suggestion
```

ansible/roles/images/tasks/load.yaml
```diff
@@ -9,12 +9,22 @@
...
-    executable: /bin/bash
-    cmd: ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }})
+- block:
+    - name: import images from files in the local images cache # noqa no-changed-when
+      shell:
+        executable: /bin/bash
+        cmd: |
+          ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
+          if [[ $? -ne 0 ]]; then
+            {{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }};
```
dkoshkin, line 19: Can we flip the statements, `mindthegap` is where we are moving towards.
The way its there currently, we would be decompressing every `item` before failing.
```suggestion
          ctr --address {{ containerd_cri_socket }} --namespace k8s.io images import <(gzip --decompress --stdout {{ item }});
          if [[ $? -ne 0 ]]; then
            {{ mindthegap_binary_location_on_remote }} import image-bundle --image-bundle={{ item }};
```

faiq: sure we can flip it

ansible/roles/offline/tasks/upload.yaml
faiq: remote copy doesn't seem to preserve the mode.

ansible/roles/images/tasks/load.yaml
faiq: It won't be present if we don't set an override with `images_local_bundle_dir`

make/images.mk
```diff
@@ -64,8 +67,12 @@ $(ARTIFACTS_DIR)/images:
 # TODO(jkoelker) UnPHONYify these targets
 .PHONY: download-images-bundle
 download-images-bundle: $(ARTIFACTS_DIR)/images
+ifeq ($(K8S_MINOR_LT_24),true)
 	curl -o $(ARTIFACTS_DIR)/images/$(DEFAULT_KUBERNETES_VERSION_SEMVER)_images$(bundle_suffix).tar.gz -fsSL https://$(AIRGAPPED_BUNDLE_URL_PREFIX)/airgapped/kubernetes-images/$(DEFAULT_KUBERNETES_VERSION_SEMVER)_images$(bundle_suffix).tar.gz
-
+else
+	export bundle_suffix="-fips"
+	curl -o $(ARTIFACTS_DIR)/images/kubernetes-images-$(DEFAULT_KUBERNETES_VERSION_SEMVER).d2iq.1$(bundle_suffix).tar.gz -fsSL https://$(AIRGAPPED_BUNDLE_URL_PREFIX)/airgapped/kubernetes-images/kubernetes-images-$(DEFAULT_KUBERNETES_VERSION_SEMVER).d2iq.1$(bundle_suffix).tar.gz
```
dkoshkin, line 74:
```
kubernetes-images-$(DEFAULT_KUBERNETES_VERSION_SEMVER)-.d2iq.1$(bundle_suffix).tar
```

dkoshkin (changes requested):

make/images.mk
faiq: ah, i keep forgetting we dropped the gz

ansible/roles/images/tasks/load.yaml
faiq: names are now -fips for images. 

good luck to whoever is changing containerd and os-packages

make/images.mk
```diff
@@ -148,7 +154,7 @@ build-%:
 		bundle_suffix=_fips \
 		download-os-packages-bundle
 	$(MAKE) pip-packages-artifacts
-	$(MAKE) bundle_suffix=_fips download-images-bundle
+	$(MAKE) download-images-bundle bundle_suffix=$$( if [ $$(echo "$(DEFAULT_KUBERNETES_VERSION_SEMVER)" | cut -f2 -d.) -lt 24 ];then echo "_fips"; else echo "-fips";fi )
```
dkoshkin, line 157: Is possible to reuse `K8S_MINOR_LT_24`?

dkoshkin (approved):

PR #482, opened 2022-08-29
feat: install nvidia driver using runfile

faiq:

Starting work on installing drivers with runfiles 

```release-note

```


make/images.mk
```diff
@@ -161,6 +173,24 @@ build-%:
...
+		os_distribution_major_version=$(call major_version,$(call version,$*)) \
+		os_distribution_arch=x86_64 \
+		download-os-packages-bundle
+	$(MAKE) download-nvidia-runfile
+	$(MAKE) pip-packages-artifacts
+	$(MAKE) download-images-bundle
+	$(MAKE) devkit.run WHAT="make build-$* \
+		BUILD_DRY_RUN=${BUILD_DRY_RUN} \
+		VERBOSITY=$(VERBOSITY) \
+		ADDITIONAL_ARGS="--instance-type=g4dn.2xlarge$(if $(ADDITIONAL_ARGS),$(SPACE)$(ADDITIONAL_ARGS))" \
```
faiq, line 191: this instance is what kaptain uses for their testing. we should do also test on this regularly. we've seen issues with this in the past 

https://jira.d2iq.com/browse/D2IQ-88483

ansible/roles/gpu/tasks/nvidia-gpu.yaml
```diff
@@ -2,18 +2,115 @@
...
+    - ansible_os_family == "RedHat"
+
+- name: regenerate the kernel initramfs
+  command: update-initramfs -u
+  become: yes
+  when: ansible_os_family == "Debian"
+
+# to pick up the changes to unload nouveau
+# some hardware like aws g4dn.2xlarge require this
+- name: unconditionally reboot the machine with all defaults
```
faiq, line 45: solves https://jira.d2iq.com/browse/D2IQ-88483

ansible/roles/gpu/defaults/main.yaml
```diff
@@ -1,50 +0,0 @@
----
-# see man lspci (-d)
-nvidia_device_query:
-  - vendor: "10de"
-    class: "0302"
-  - vendor: "10de"
-    class: "0300"
-vault_repo_baseurl: "http://vault.centos.org/{{ os_release_file['content'] |b64decode | regex_search('\\d+\\.\\d+\\.\\d+') }}/os/$basearch/"
```
supershal, line 8: These variables are still used in `ansible/roles/gpu/tasks/nvidia-gpu-CentOS7.yaml`

ansible/roles/gpu/tasks/nvidia-gpu.yaml
```diff
@@ -2,18 +2,115 @@
...
+  shell: |
+    "{{ nvidia_remote_bundle_path }}/{{ nvidia_runfile_installer }}" -x -s
+  args:
+    chdir: /opt/dkp/nvidia
+  become: yes
+
+- name: install driver
+  shell:
+    ./nvidia-installer -s \
+      --log-file-name=$(pwd)/nvidia-installer.log || true
```
supershal, line 73: should not ignore exit code for driver installation failures

ansible/roles/gpu/defaults/main.yaml
faiq: it seems like thats for unsupported centos versions https://github.com/mesosphere/konvoy-image-builder/blob/main/ansible/roles/gpu/tasks/nvidia-gpu-CentOS7.yaml#L31

let me confirm

ansible/roles/gpu/tasks/nvidia-gpu.yaml
faiq: the test we're using is `nvidia-smi` to determine whether or not the installation works let me try removing and see what it does. the installer can sometimes give a false negative which is why i added the `|| true` there

ansible/roles/gpu/tasks/nvidia-gpu.yaml
```diff
@@ -2,18 +2,111 @@
...
+# some hardware like aws g4dn.2xlarge require this
+- name: unconditionally reboot the machine with all defaults
+  ansible.builtin.reboot:
+  when: sysprep
+
+# Download and install gpu runfile
+- name: create nvidia install directory
+  file:
+    path: /opt/dkp/nvidia
+    state: directory
```
supershal, line 49: ansible re-runs gives error when `/opt/dkp/nvidia` already exists. Delete `/opt/dkp/nvidia` to delete old installations first or add file exists check.

ansible/roles/gpu/tasks/nvidia-gpu.yaml
```diff
@@ -2,18 +2,111 @@
...
+# to pick up the changes to unload nouveau
+# some hardware like aws g4dn.2xlarge require this
+- name: unconditionally reboot the machine with all defaults
+  ansible.builtin.reboot:
+  when: sysprep
+
+# Download and install gpu runfile
+- name: create nvidia install directory
+  file:
+    path: /opt/dkp/nvidia
```
supershal, line 48: use `nvidia_remote_bundle_path` here and wherever applicable.

ansible/roles/gpu/tasks/nvidia-gpu.yaml
```diff
@@ -2,18 +2,111 @@
...
+- name: Write location of library files
+  lineinfile:
+    path: /etc/ld.so.conf.d/nvidia.conf
+    line: '/usr/lib64'
+    insertbefore: BOF
+
+- name: run nvidia-smi
+  shell:
+    executable: /bin/bash
+    cmd: nvidia-smi
```
supershal, line 92: chdir to `{{ nvidia_remote_bundle_path }}/NVIDIA-Linux-x86_64-{{ nvidia_driver_version }}"`

faiq: it should add it to the path on install.

supershal, line 92:
```suggestion
    cmd: "{{ nvidia_remote_bundle_path }}/NVIDIA-Linux-x86_64-{{ nvidia_driver_version }}/nvidia-smi"
```

supershal (approved):
Great work. 🎉

alejandroEsc (approved):
not sure I understand all of it but looks good to me. One comment I would make is the naming for certain places has inconsistent nvidia naming, not a huge issue but something to consider in the future.

ansible/roles/gpu/tasks/nvidia-gpu.yaml
dkoshkin: Will the machine also reboot for preprovisioned, or is `sysprep==false` in that case?

overrides/offline-nvidia.yaml
```diff
@@ -0,0 +1,7 @@
+nvidia_runfile_local_file: "{{ playbook_dir}}/../artifacts/{{ nvidia_runfile_installer }}"
```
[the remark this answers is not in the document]
faiq: added

ansible/roles/gpu/tasks/nvidia-gpu.yaml
faiq: it's false. 

however, they might need to reboot it on their own depending on OS if they were previously using nouveu drivers

ansible/roles/gpu/tasks/nvidia-gpu-Ubuntu20.yaml
```diff
@@ -2,80 +2,12 @@
 # Implements the NVIDIA Driver Installation Quickstart Guide for Ubuntu LTS - see
 # https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html#ubuntu-lts
 
-- name: install kernel headers
+- name: install runfile prereqs
   apt:
-    name: linux-headers-{{ ansible_kernel }}
+    name:
+      - linux-headers-{{ ansible_kernel }}
```
faiq, line 8:
```
$ apt-cache depends build-essential
build-essential
 |Depends: libc6-dev
  Depends: <libc-dev>
    libc6-dev
  Depends: gcc
  Depends: g++
  Depends: make
    make-guile
  Depends: dpkg-dev
```

faiq: makes it a little cleaner :)

fatz:
just want to make clear that this patch deleted the vault functionality to make sure staying with the centos minor version kernel.

That means whenever RedHat releases a 7.10 release we'll either make a half 7.9 half 7.10 image as we'll install 7.10 patch in a 7.9 image or we'll simply have to immediately drop support for 7.9 to follow 7.10. 

But the minor version is hard coded into every KIB build. So @faiq I really really suggest keeping this backward compatibility as we render previous KIB versions unusable as soon as there is a new RHEL7/CentOS7 release 

CentOS moves almost instant the kernel packages of 7.9 into vault once a new release is there.

This is the reason for the vault code as it drops the dependency on CentOS releases and timeline

PR #493, opened 2022-09-09
fix: Configure containerd runc plugin options with systemd cgroup driver

dlipovetsky:
konvoy-image-builder installs kubelet as a systemd service, and kubelet therefore uses the systemd cgroup driver.

The container runtime must use the same cgroup driver.

> Note: In v1.22, if the user is not setting the cgroupDriver field
> under KubeletConfiguration, kubeadm will default it to systemd.
> -- https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/

```release-note

```


ansible/roles/config/templates/config.toml.tmpl
```diff
@@ -96,6 +95,8 @@ imports = ["/etc/containerd/conf.d/*.toml"]
           runtime_engine = ""
           runtime_root = ""
           privileged_without_host_devices = false
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            systemd_cgroup = true
```
dkoshkin, line 99: Is this value dependent on the Kubernetes version?

dlipovetsky:
@faiq raised a good question about _where_ in the containerd configuration this PR sets the cgroup driver.

The "global" setting has been deprecated for a while: https://github.com/containerd/containerd/blob/f4095a6876c62dd4cd5e3749d2435b2daa9eaff8/pkg/cri/config/config.go#L259-L262

So the right place to set it appears to be in the CRI runtime plugin

ansible/roles/config/templates/config.toml.tmpl
dlipovetsky: Yes, but as of v1.21, kubeadm `init` configures kubelet to use the systemd cgroup driver, and as of v1.22, _all_ kubeadm commands configure kubelet to use it.

Please see https://github.com/kubernetes/kubeadm/issues/2376#issuecomment-788252505

dlipovetsky:
Looks like setting the configuration here only works for runtime type `io.containerd.runc.v2`. We use `io.containerd.runc.v1`, which means we need to change the configuration at the original, global location.

ansible/roles/config/templates/config.toml.tmpl
```diff
@@ -92,14 +91,17 @@ imports = ["/etc/containerd/conf.d/*.toml"]
         privileged_without_host_devices = false
       [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
         [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
-          runtime_type = "io.containerd.runc.v1"
+          runtime_type = "io.containerd.runc.v2"
           runtime_engine = ""
           runtime_root = ""
           privileged_without_host_devices = false
+          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+            systemd_cgroup = true
```
faiq, line 99:
```suggestion
            SystemdCgroup = true
```

faiq: I think we want this, based on some testing

dlipovetsky: Confirmed independently. I ran `kubeadm init`, then checked the configuration of the various control-plane containers that run in Static Pods:

Using `systemd_cgroup`:
```shell
# crictl ps -q | xargs -n1 crictl inspect | grep systemd_cgroup
```

Using `SystemdCgroup`
```shell
# crictl ps -q | xargs -n1 crictl inspect | grep systemd_cgroup
      "systemd_cgroup": true
      "systemd_cgroup": true
      "systemd_cgroup": true
      "systemd_cgroup": true
      "systemd_cgroup": true
```

faiq:
There's some flatcar specific changes I'd like to take a look at before merging as well

faiq (approved):
No customer is using flatcar with dkp 2.2, which introduced disabling cgroups v2. So we can safely ignore the case where a flatcar configured with cgroupsv1 needs to be upgraded to dkp 2.4

PR #496, opened 2022-09-13
WIP: use hcl files instead of json

fatz:
We're still relying on packer JSON templates which will still be supported but don't receive any new packer features. HCL will make it way easier for us maintaining multiple Distros for a single deployment.

* https://jira.d2iq.com/browse/D2IQ-92870

This is still WIP and not ready to use

```release-note

```


fatz:
started a new branch at #507 which replaces this work

PR #507, opened 2022-09-20
feat: HCL instead of JSON packer templates 

fatz:
We're still relying on packer JSON templates which will still be supported but don't receive any new packer features. HCL will make it way easier for us maintaining multiple Distros for a single deployment which allows us to support things like iso based building on vsphere

* https://jira.d2iq.com/browse/D2IQ-92870

- no change expected. Just the packer format changed from json to hcl.
- plugins get installed in an additional step
- goss is used without plugin reference as it currently does not support the plugin registry format

```release-note

```


fatz:
hey @supershal I actually got some trouble with the vsphere stuff here.

when using a base image from our vsphere cluster i get 
```
Build 'vsphere-clone.kib_image' errored after 13 seconds 471 milliseconds: error saving temporary keypair in the vm: not possible to save temporary public key: this VM lacks a vApp configuration and cannot have vApp properties set on it
```

maybe I chose the wrong image please point me to a working one... otherwise I wonder how we made this work in the past...

fatz:
@dkoshkin could you please have a look at this again.

I was able to add goss again by just leaving out the plugin registry part and rely on the old fashion install method like with json.

supershal:
getting following error in
```
* AMIName should only contain alphanumeric characters, parentheses (()), square brackets ([]), spaces ( ), periods (.), slashes (/), dashes (-), single quotes ('), at-signs (@), or underscores(_). You can use the `clean_resource_name` template filter to automatically clean your ami name.
```
`clean_resource_name` is not implemented in HCL2. recommended function in HCL2 is [regex_replace](https://www.packer.io/docs/templates/hcl_templates/functions/string/regex_replace)

We need to create regexp for each AWS, GCP and Azure: https://www.packer.io/docs/templates/legacy_json_templates/engine#clean_resource_name

supershal:
The packer JSON files were go templates which mostly supported `--dry-run` functionality. When `--dry-run` is enabled, packer deletes AMI, OVA, Azure image, GCP image once build is successful. This prevents unnecessary AMI created when we run e2e builds and we dont need separate process to clean up resources.

The aws packer manifest also uses templating to determine if packer should use source ami or ami filters based on user inputs.

These functionalities are not moved HCL manifests. 
Ideally I prefer to not convert these files into Go templates as we dont use it extensively and most of the templating can be replaced by go code + packer variables (It may not be possible to replace some of current templating blocks with packer variables). Keeping these files as `.hcl` would also make it easy to test locally for small changes. 

We should check into how can we replace current go templating.

supershal (changes requested):

pkg/packer/manifests/gcp/packer.pkr.hcl
```diff
@@ -0,0 +1,363 @@
+packer {
+  required_plugins {
+    googlecompute = {
+      version = ">= 1.0.11"
```
supershal, line 4: should we keep plugin version constraints in one place in `Dockerfile.devkit` that way we wont mismatch version requirements and make sure that required plugins versions are available for air-gapped installations.

pkg/packer/manifests/vsphere/packer.json.tmpl
```diff
@@ -1,195 +0,0 @@
...
-        "kubernetes_typed_version": "{{user `kubernetes_typed_version`}}",
-        "os_name": "{{user `os_display_name`}}",
-        "vsphere_guest_os_type": "{{user `vsphere_guest_os_type`}}"
-      },
-      "output": "{{user `manifest_output`}}",
-      "strip_path": true
-    }
-    ((- if .DryRun ))
-    ,{
-      "type": "shell-local",
```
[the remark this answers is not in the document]
fatz: could you elaborate what this is about? Packer is destroying the VM by itself. What is the reason exposing credentials into the VM?

fatz:
> The packer JSON files were go templates which mostly supported `--dry-run` functionality. When `--dry-run` is enabled, packer deletes AMI, OVA, Azure image, GCP image once build is successful. This prevents unnecessary AMI created when we run e2e builds and we dont need separate process to clean up resources.

yeah good point... just forgot to add this again... Commit is on its way

> The aws packer manifest also uses templating to determine if packer should use source ami or ami filters based on user inputs.

this is fully implemented in HCL. See `data` block and `local.source_ami`

> These functionalities are not moved HCL manifests. Ideally I prefer to not convert these files into Go templates as we dont use it extensively and most of the templating can be replaced by go code + packer variables (It may not be possible to replace some of current templating blocks with packer variables). Keeping these files as `.hcl` would also make it easy to test locally for small changes.
> 
> We should check into how can we replace current go templating.

there is no need of templating anymore. But deleting the whole rendering logic would have blown up this PR too much. Essentially its not needed anymore. Such a cleanup could be done in another PR

pkg/packer/manifests/gcp/packer.pkr.hcl
fatz: sync the versions of devkit and hcl...

However when writing HCL this specifies a minimum version which is in this case the one this was developed against. 

Furthermore adding the devkit downloads is only be done because of goss... ideally we will use `packer init` in the future which makes specifying it in the container useless

supershal: makes sense. thank you.

supershal (approved):

faiq (approved):

PR #508, opened 2022-09-21
feat: configure fips for rhel 8

faiq:

* https://d2iq.atlassian.net/browse/D2IQ-93017

```release-note

```


ansible/roles/fips/tasks/main.yaml
```diff
@@ -0,0 +1,14 @@
+- name: "RHEL 7 FIPS setup"
```
dkoshkin, line 1: Please add a link here too

ansible/roles/fips/tasks/redhat-7.yaml
```diff
@@ -0,0 +1,71 @@
...
+    path: /etc/sysconfig/prelink
+    regexp: '^PRELINKING=yes'
+    replace: '^PRELINKING=no'
+  when:
+    - sysfips is changed
+
+- name: "redhat 7 fips: install dracut fips"
+  yum:
+    name:
+      - dracut-fips
```
dkoshkin, line 21: This package is not in the bundle, maybe we should only do this for >RHEL 8?

ansible/roles/fips/tasks/redhat-8.yaml
```diff
@@ -0,0 +1,19 @@
...
+- name: "redhat 8 fips: run fips enable command"
+  command: fips-mode-setup --enable
+  when:
+    - sysfips is changed
+
+- name: "redhat 8 fips: reboot to pick up fips changes"
+  ansible.builtin.reboot:
+  when:
+    - sysfips is changed
+    - fips_reboot
```
dkoshkin, line 19: I don't think it makes sense to change the FIPS mode but not reboot. IMO it should be all or nothing.

overrides/fips.yaml
```diff
@@ -4,6 +4,8 @@ k8s_image_registry: docker.io/mesosphere
 fips:
   enabled: true
```
dkoshkin, line 5: Or something like that
```suggestion
fips:
  enabled: true
  # when set to true and FIPS mode is not already enabled, FIPS mode will be enabled and the machine will be rebooted
  # only works on RHEL 8 
  configure: false
```

ansible/roles/fips/tasks/redhat-7.yaml
faiq: we're doing it for rhel 7 thats it. i still haven't worked on testing this bit

overrides/fips.yaml
faiq: yes actually i love this. 

we should guard all the tasks in `fips` role  to only run if `configure == true`

ansible/roles/fips/tasks/redhat-7.yaml
dkoshkin: I meant not have this codepath at all for RHEL 7. Only do https://github.com/mesosphere/konvoy-image-builder/pull/508/files#diff-5f48c283af385dc86e531a1d0821000e9fee1c77cc9755ee4cc4c3f2c5b0eb76R1 for RHEL 8.

overrides/fips.yaml
dkoshkin: Looks like you forgot to make this change too.
You will also need to override this configuration in KIB's CI.

faiq: right now im keeping this bit locally and for CI I'm planning on doing it in an extra override passed  to here.

ansible/roles/fips/tasks/redhat-7.yaml
faiq: removed

ansible/roles/fips/tasks/redhat-8.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+- name: "redhat 8 fips: run fips enable command"
+  command: fips-mode-setup --enable
+  when:
+    - sysfips is changed
+
+- name: "redhat 8 fips: reboot to pick up fips changes"
+  ansible.builtin.reboot:
+  when:
+    - sysfips is changed
```
supershal, line 18: nit: `sysfips is changed` status check is valid. wdyt adding another condition to check success status of `fips-mode-setup --enable` task

supershal: please disregard above. the reboot task will proceed only if `fips-mode-setup --enable` is successful. no need to unnecessary check for its success.

supershal (approved):

overrides/fips.yaml
faiq: added these to CI. and made the default false

dkoshkin (approved):

ansible/roles/fips/tasks/main.yaml
```diff
@@ -0,0 +1,8 @@
+# https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening#switching-the-system-to-fips-mode_using-the-system-wide-cryptographic-policies
```
faiq, line 1:
```suggestion
# see: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening#switching-the-system-to-fips-mode_using-the-system-wide-cryptographic-policies
```

PR #515, opened 2022-09-28
fix: remove fips.configure from a user facing override

dkoshkin:
This is tested in a limit set of OSs, lets not expose it yet to users.
The value is already being set to `false` in https://github.com/mesosphere/konvoy-image-builder/blob/d226e7aeec35229c31f510a9862daecda6595c10/ansible/group_vars/all/system.yaml#L22-L24

```release-note

```

### DIFF
--- a/overrides/fips.yaml
+++ b/overrides/fips.yaml
@@ -3,7 +3,6 @@ k8s_image_registry: docker.io/mesosphere
 
 fips:
   enabled: true
-  configure: false
 
 build_name_extra: -fips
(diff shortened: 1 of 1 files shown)


msdolbey (approved):
👍 
nothing to see here

cneth:
How would a customer ensure KIB images boot into fips mode without this setting?   I believe 'strict fips' mode requires that machines always boot with the fips=1 flag so that all crypto material that might be generated on 'startup' uses FIPS algo's...

PR #532, opened 2022-10-12
fix: add retries when installing SUSE packages 

supershal:
- attempt to fix E2E flakes because of following errors. Some `zypper` tasks missing retries. This PR sets same retry and delay for all `zypper` tasks.

```
sles-15-nvidia: TASK [containerd : remove preinstalled containerd] *****************************
      sles-15-nvidia: fatal: [default]: FAILED! => {"changed": false, "cmd": ["/usr/bin/zypper", "--quiet", "--non-interactive", "--xmlout", "--disable-repositories", "search", "--type", "package", "--match-exact", "--details", "--installed-only", "containerd"], "msg": "Zypper run command failed with return code 7.", "rc": 7, "stderr": "", "stderr_lines": [], "stdout": "<?xml version='1.0'?>\n<stream>\n<message type=\"error\">System management is locked by the application with pid 4463 (zypper).\nClose this application before trying again.</message>\n</stream>\n", "stdout_lines": ["<?xml version='1.0'?>", "<stream>", "<message type=\"error\">System management is locked by the application with pid 4463 (zypper).", "Close this application before trying again.</message>", "</stream>"]}
```

I looked up SUSE forums for this. only suggestion i found is to manually `kill` the running task. 
The delay(1 minute) and retry attempts (15) are large, but keeping them to match with existing tasks. These can be tuned once we are sure that this eliminates flakes we are seeing in the CI.

### DIFF
--- a/ansible/roles/containerd/tasks/suse.yaml
+++ b/ansible/roles/containerd/tasks/suse.yaml
@@ -4,6 +4,10 @@
   zypper:
     name: containerd
     state: absent
+  register: result
+  until: result is success
+  retries: 15
+  delay: 60
 
 - name: remove versionlock for containerd
   command: zypper removelock containerd.io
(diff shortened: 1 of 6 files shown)


jkoelker (approved):

dkoshkin (approved):
Lets try it

PR #534, opened 2022-10-17
fix: add kuberentes version to azure gallery image name

supershal:
- Create azure gallery image per kubernetes version and image flavors. ex. `dkp-ubuntu-2004-release-1.24.6`, `dkp-ubuntu-2004-release-1.24.6-fips.0`,  `dkp-ubuntu-2004-release-1.24.6-nvidia`

New VMw can be created using `latest` version of the gallery image id using full gallery id:
`/subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/dkp/providers/Microsoft.Compute/galleries/dkp/images/dkp-ubuntu-2004-release-{{.K8sVersion}}/versions/latest`

This helps Azure e2e tests to always pick up `latest` image for a given kubernetes version.

- Set default replication of the azure images to `westus` to match with CAPZ defaults


cmd/konvoy-image/cmd/azure.go
```diff
@@ -122,8 +122,8 @@ func addAzureArgs(fs *flag.FlagSet, azure *app.AzureArgs) {
 	fs.StringArrayVar(
 		&azure.GalleryImageLocations,
 		"gallery-image-locations",
-		[]string{},
-		"a list of locatins to publish the image (default same as `location`)",
+		[]string{"westus"},
```
supershal, line 125: setting to `westus` to match with default location used by CAPZ

pkg/app/config.go
```diff
@@ -487,11 +489,23 @@ func MergeAzureUserArgs(config Config, azureArgs *AzureArgs) error {
...
+		return fmt.Errorf("unable to get full kubernetes version from config: %w", err)
+	}
 	galleryImageName := azureArgs.GalleryImageName
 	if galleryImageName == "" {
-		galleryImageName = fmt.Sprintf("dkp-%s", BuildName(config))
+		galleryImageName = fmt.Sprintf("dkp-%s-%s", BuildName(config), fullKuberenetesVersion)
 	}
 
+	galleryRegex := regexp.MustCompile(azureaGlleryImageNameRegex)
+	galleryImageName = galleryRegex.ReplaceAllString(galleryImageName, "-")
```
supershal, line 508: only alphabets, numbers, `. _ -` allowed.  `fips+0` breaks this pattern. so sanitizing the gallery image name.

pkg/azure/azure.go
```diff
@@ -211,43 +211,41 @@ func EnsureImageDescriptions(
 		return fmt.Errorf("failed to obtain a credential: %w", err)
 	}
 
-	for _, location := range locations {
```
supershal, line 214: azure resource, gallery and gallery image can be created with only one location. image can be replicated in multiple locations (regions). The replication of the images will be handled by packer.

pkg/app/config.go
```diff
@@ -510,7 +524,16 @@ func MergeAzureUserArgs(config Config, azureArgs *AzureArgs) error {
 		return fmt.Errorf("failed to set %s: %w", PackerAzureGalleryImagePublisherPath, err)
 	}
 
-	if err := config.Set(PackerAzureGalleryImageSKU, azureArgs.GalleryImageSKU); err != nil {
+	galleryImageSKU := azureArgs.GalleryImageSKU
+	if galleryImageSKU == "" {
```
alejandroEsc, line 528: would it be a terrible thing to require this instead of setting it to the gallery name?

alejandroEsc (commented):
Just one question, looking great!

pkg/app/config.go
supershal: The earlier implementation defaults to `distribution_version` if its not provided. If we make it `required` we have to make sure to validate and fail earlier. The validation would require call to Azure and make sure the SKU does not exists within given `subscription/resourcegroup/gallery/gallery_image_id`. 
IMHO, we should remove/deprecate this flag in the CLI and not ask gallery_name, sku etc to simplify provision. 
The azure documentation around gallery offer, publisher and sku and gallery image URN are not clear enough. Instead we can make sure that KIB ensures unique URN for the image.

faiq (approved):

PR #544, opened 2022-10-24
build: enable Nvidia offline for rhel 8.6

alejandroEsc:
This allows for the makefile targets to build rhel 8.6 offline gpu and also fixes building for rhel.

* https://jira.d2iq.com/browse/D2IQ-92651

```release-note

```

### DIFF
--- a/make/ci.mk
+++ b/make/ci.mk
@@ -55,6 +55,8 @@ e2e.build.rhel-8.4-offline-fips: rhel84-fips-offline infra.aws.destroy
 
 e2e.build.rhel-8.4-offline-nvidia: rhel84-offline-nvidia infra.aws.destroy
 
+e2e.build.rhel-8.6-offline-nvidia: rhel86-offline-nvidia infra.aws.destroy
+
 e2e.build.rhel-8.6-offline-fips: rhel86-fips-offline infra.aws.destroy
 
 e2e.build.ubuntu-18: ubuntu18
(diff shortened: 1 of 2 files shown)


ansible/roles/gpu/tasks/nvidia-gpu-RHEL.yaml
```diff
@@ -9,12 +9,26 @@
     enablerepo: "{{ 'offline' if offline_mode_enabled else '' }}"
     disablerepo: "{{ '*' if offline_mode_enabled else '' }}"
 
-- name: install additional Nvidia runfile prerequisites
+- name: find installed Nvidia RPMs
```
faiq, line 12: can we see if we need these or if including the missing library packages was enough.

alejandroEsc: I think it's cleaner to have this approach anyway, but yeah sure I can test this tomorrow morning.

alejandroEsc: Removed the commit with this task change as it is not necessary and tested it out to work.

dkoshkin (approved):

alejandroEsc:
Working as expected: https://teamcity.mesosphere.io/buildConfiguration/ClosedSource_KonvoyImageBuilder_BuildAmiRhel86offlineNvidia/3667505?buildTab=log&expandAll=true&focusLine=2977

faiq (approved):

PR #548, opened 2022-10-25
fix: always set feature flag

faiq:
**What problem does this PR solve?**:
https://d2iq.atlassian.net/browse/D2IQ-93868

upstream seems to always sets these flags https://github.com/kubernetes-sigs/image-builder/blob/02df45969409c7f18f2cf7e63b70c4dc0e2fa442/images/capi/ansible/roles/providers/tasks/main.yml#L105

i started vm with ubuntu 20.04 and ran cloud-init version 

```bash
root@ip-10-0-122-187:/var/snap/amazon-ssm-agent/6312# cloud-init --version
/usr/bin/cloud-init 22.3.4-0ubuntu1~20.04.1
```

so i also updated the regex and the version comparision command as well.

See this conversation in the kubernetes slack for why we included this version check in the first place https://kubernetes.slack.com/archives/CD6U2V71N/p1629737897163400 

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-NUMBER

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #550, opened 2022-10-26
fix: set feature flag for cloud-init to ignore errors

fatz:
- feature flags for ubuntu 20.04 not set due to regex and version missmatch. 
- on centos 7.9 no featureflags set due to cloud-init sending version string on stderr 
- supersceedes #548 

* https://jira.d2iq.com/browse/D2IQ-93868

```release-note
fix: cloud-init feature flags where not set on ubuntu 20.04 leading to images not being usable by CAPI
```

### DIFF
--- a/ansible/molecule/default/molecule.yml
+++ b/ansible/molecule/default/molecule.yml
@@ -10,12 +10,6 @@ platforms:
     region: us-east-1
     instance_type: t3.small
     ssh_user: centos
-  - name: konvoyimage-centos8.3-${USER:-ci}-${HOSTNAME:-local}
-    image_search_name: "CentOS 8.3.2011 x86_64"
-    image_search_owner: "125523088429"
-    region: us-east-1
-    instance_type: t3.small
-    ssh_user: centos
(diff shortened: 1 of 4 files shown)


fatz:
so when executing molecule tests centos-7 is failing
```
fatal: [konvoyimage-centos7.9-julferts-local]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'epel_centos_7
_rpm_gpg_key' is undefined\n\nThe error appears to be in '/Users/julferts/golang/src/github.com/mesosphere/konvoy-image-builder/ansible/roles/repo/tasks
/redhat.yaml': line 44, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- n
ame: add epel gpg key for centos 7\n  ^ here\n"}
```

this seems to be related to `main` branch as #551 works absolutely fine

fatz:
the fatal above is due to missing defaults in repo role. #482 removed defaults from GPU module as that role was included by defaults its defaults where used by the repo role. The last commit makes it clear and removes external role dependencies of the repo role

PR #554, opened 2022-10-28
chore: use konvoy image wrapper for  build targets 

supershal:
- Use konvoy image wrapper binary as default for all CI build targets (TC configuration does not needs to be changed)
  - This will make the CI runs same as how production users would use KIB
- Remove using `konvoy-image` binary for running image build targets
- Replace direct use of `devkit` container to run `konvoy-image` binary. 
- Refactor `wrapper.go` (first pass) to better organize code
- start using `make build-wrapper` instead of `make build` for development. All dependency required for building wrapper will be handled by Makefile targets automatically. 
- **Enable running konvoy-image command on Mac without Sudo access, hacking Makefile, changing file permissions,  propagating env variables to sudo etc**

**feature enhancements:**
-  remove direct use of `devkit` container. use `devkit` container only as a base for building konvoy-image docker image.
- make `devkit` container "static" and upload to DockerHub instead of building it all the time.
- refactor `wrapper.go` file one more time to reduce boilerplate code
- Move all build target to use `goreleaser` to make sure same mechanism is used for development, CI and creating releases. 

```release-note

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ FROM alpine:3.15.4
 
 ENV ANSIBLE_PATH=/usr
 ENV PYTHON_PATH=/usr
+ENV PACKER_PLUGIN_PATH=/opt/packer/plugins
 
 COPY requirements.txt /tmp/
 # NOTE(jkoelker) Ignore "Pin versions in [pip | apk add]"
@@ -28,7 +29,7 @@ COPY --from=devkit /usr/local/bin/mindthegap /usr/local/bin/
 COPY --from=devkit /usr/local/bin/packer /usr/local/bin/
 COPY --from=devkit /usr/local/bin/packer-provisioner-goss /usr/local/bin/
 COPY --from=devkit /usr/local/bin/govc /usr/local/bin/
-COPY --from=devkit /root/.config/packer/plugins/ /root/.config/packer/plugins/
+COPY --from=devkit /root/.config/packer/plugins/ ${PACKER_PLUGIN_PATH}
 COPY bin/konvoy-image /usr/local/bin
 COPY images /root/images
 COPY ansible /root/ansible
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ COPY --from=devkit /usr/local/bin/goss /usr/local/bin/
 COPY --from=devkit /usr/local/bin/mindthegap /usr/local/bin/
 COPY --from=devkit /usr/local/bin/packer /usr/local/bin/
 COPY --from=devkit /usr/local/bin/packer-provisioner-goss /usr/local/bin/
+COPY --from=devkit /usr/local/bin/govc /usr/local/bin/
 COPY --from=devkit /root/.config/packer/plugins/ /root/.config/packer/plugins/
 COPY bin/konvoy-image /usr/local/bin
 COPY images /root/images
--- a/Dockerfile.devkit
+++ b/Dockerfile.devkit
@@ -110,14 +110,13 @@ COPY --from=docker /usr/local/bin/docker /usr/local/bin/
 COPY --from=govc /govc /usr/local/bin/
 COPY --from=builder /tools /usr/local/bin
 
-
 RUN getent group "${GROUP_ID}" > /dev/null 2>&1 \
     || addgroup -S -g "${GROUP_ID}" "${GROUP_NAME}"
 
 # NOTE(jkoelker) Ignore "Multiple consecutive `RUN`"
 # hadolint ignore=DL3059
 RUN getent passwd "${USER_ID}" > /dev/null 2>&1 \
-    || adduser -D -u "${USER_ID}" -G "${GROUP_NAME}" "${USER_NAME}"
+    || adduser -D -u "${USER_ID}" -g "${GROUP_ID}" "${USER_NAME}"
 
 # NOTE(jkoelker) Ignore "A && B || C is not if-then-else"
 # NOTE(jkoelker) Ignore "Multiple consecutive `RUN`"
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,8 @@ INVENTORY_FILE ?= $(REPO_ROOT_DIR)/inventory.yaml
 COMMA:=,
 
 export CGO_ENABLED=0
+export GO_VERSION := 1.19 #$(shell grep -oP '^(go )\K(\d+.\d+)$$' go.mod)
+GOLANG_IMAGE := golang:$(GO_VERSION)
 
 export CI ?= no
 ifeq ($(CI),yes)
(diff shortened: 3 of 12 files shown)


supershal:
TC configuration for build changed to use `make build-wrapper`
https://teamcity.mesosphere.io/admin/editBuildRunners.html?id=buildType:ClosedSource_KonvoyImageBuilder_Build

Dockerfile
```diff
@@ -27,7 +28,8 @@ COPY --from=devkit /usr/local/bin/goss /usr/local/bin/
 COPY --from=devkit /usr/local/bin/mindthegap /usr/local/bin/
 COPY --from=devkit /usr/local/bin/packer /usr/local/bin/
 COPY --from=devkit /usr/local/bin/packer-provisioner-goss /usr/local/bin/
-COPY --from=devkit /root/.config/packer/plugins/ /root/.config/packer/plugins/
```
supershal, line 30: when running kib wrapper as non-root user, the packer could not find and load the plugins and defaults to in-built version of the plugins. so moving the plugin to `/opt/packer/plugins` on the container so that packer can find correct version of the plugins.

Dockerfile.devkit
```diff
@@ -110,14 +110,13 @@ COPY --from=docker /usr/local/bin/docker /usr/local/bin/
...
 
-
 RUN getent group "${GROUP_ID}" > /dev/null 2>&1 \
     || addgroup -S -g "${GROUP_ID}" "${GROUP_NAME}"
 
 # NOTE(jkoelker) Ignore "Multiple consecutive `RUN`"
 # hadolint ignore=DL3059
 RUN getent passwd "${USER_ID}" > /dev/null 2>&1 \
-    || adduser -D -u "${USER_ID}" -G "${GROUP_NAME}" "${USER_NAME}"
+    || adduser -D -u "${USER_ID}" -g "${GROUP_ID}" "${USER_NAME}"
```
supershal, line 119: Mac users get error when adding the user since the ID of the `staff` group on mac matches with root user on devkit container. This fix will add the user using the group id in the container.

This issue does not happen when using konvoy-image-wrapper since the wrapper code masks the groups and passwd files.

Makefile
```diff
@@ -178,7 +180,7 @@ $(DOCKER_DEVKIT_PHONY_FILE): Dockerfile.devkit
 	&& touch $(DOCKER_DEVKIT_PHONY_FILE)
 
 $(DOCKER_PHONY_FILE): $(DOCKER_DEVKIT_PHONY_FILE)
-$(DOCKER_PHONY_FILE): bin/konvoy-image
+$(DOCKER_PHONY_FILE): konvoy-image-linux
```
supershal, line 183: creating linux binary specifically so that mac users dont have to set GOOS=linux explicitly when building the kib devkit image.

Makefile
```diff
@@ -240,14 +253,18 @@ bin/konvoy-image: $(shell find $(REPO_ROOT_DIR)/pkg -type f -name '*'.tmpl)
...
+konvoy-image-linux:
+	$(MAKE) docker GOOS=linux WHAT="make bin/konvoy-image"
+
+bin/konvoy-image-wrapper: $(DOCKER_PHONY_FILE)
+bin/konvoy-image-wrapper: 
 	$(call print-target)
-	go build \
-		-ldflags="-X github.com/mesosphere/konvoy-image-builder/pkg/version.version=$(REPO_REV)" \
-		-o ./bin/konvoy-image-wrapper ./cmd/konvoy-image-wrapper/main.go
+	$(MAKE) docker WHAT="go build \
```
supershal, line 265: making sure to use the same golang version as used in go.mod file.

cmd/konvoy-image-wrapper/cmd/wrapper.go
```diff
@@ -213,9 +214,11 @@ func (r *Runner) mountFileEnv(envName string, containerPath string) error {
 	fi, err := os.Stat(filePath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return fmt.Errorf("env %s is set, but file %s does not exist", envName, filePath)
+			// Ignore if file does not exists
```
supershal, line 217: The wrapper binary is common for all the infra providers. for some providers the creds files are optional while some requires it (`gcp`). This breaks our CI where we always set all the env variables for all the builds. So adding fix to ignore if file does not exists. The infrastructure specific validation is moved to its subcommand. check `konvoy-image/cmd/gcp.go`

cmd/konvoy-image-wrapper/main.go
```diff
@@ -11,5 +11,6 @@ func main() {
 	err := cmd.NewRunner().Run(os.Args[1:])
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "error encountered: %s\n", err)
+		os.Exit(1)
```
supershal, line 14: The KIB wrapper never exited with non-zero code before.

cmd/konvoy-image-wrapper/cmd/wrapper.go
dkoshkin: That makes sense, and because [in this line](https://github.com/mesosphere/konvoy-image-builder/pull/554/files/aa5ea098dadeccdceba37e5c5611f64faabc7701#diff-d8016cb551cfda4bbd3bc9124263f9003367a4cf5a6d144b96c928faec12f3aaR204) the same file path will be mounted its ok to move it into the `cmd` package.

Please just a comment in `setGCPEnv` that we expect the mount path to equal the path on the host.

supershal: thank you. sounds good.

Makefile
```diff
@@ -383,6 +403,7 @@ define print-target
     @printf "Executing target: \033[36m$@\033[0m\n"
 endef
 
+release-bundle-GOOS: cmd/konvoy-image-wrapper/image/konvoy-image-builder.tar.gz
```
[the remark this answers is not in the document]
supershal: I moved it there because i just wanted to create bundle for only darwin.
However while fixing it, i realized that this target is not used and duplicate of `release-snapshot` target which uses goreleaser to create release bundles. Should we remove this target instead of fixing it and rely on single mechanism that creates release bundles using goreleaser. wdyt?

faiq (approved):

supershal:
while testing `release:` target, I found that the PR broke release target. The PR introduces golang docker container to build go binary and finds go version from `go.mod` file. The `grep` options used to find go version does not work when executed from devkit container. so replace it with `sed` expression. Please check https://github.com/mesosphere/konvoy-image-builder/pull/554/commits/8d6ff3658229c84c21509ca0868d058927f39345

dkoshkin (approved):

PR #559, opened 2022-11-01
fix: Can not specify custom Packer directory 

mkarroqe:
This PR supports packer provisioning with STIG requirements by using the `remote_folder` parameter in packer to override the default `/tmp` directory.

* https://jira.d2iq.com/browse/D2IQ-93657/

- @supershal and I tested using a custom `source_ami` image in `rhel-82.yaml` where we remounted the `/tmp` directory with the `noexec` flag to match STIG requirements. We also commented out `ami_filter_name` and `ami_filter_owners`.

### DIFF
--- a/images/ami/rhel-82.yaml
+++ b/images/ami/rhel-82.yaml
@@ -1,12 +1,12 @@
 download_images: true
 
 packer:
-  # ami_filter_name: "RHEL-8.2.0*2020*-x86_64-*"
-  # ami_filter_owners: "309956199498"
+  ami_filter_name: "RHEL-8.2.0*2020*-x86_64-*"
+  ami_filter_owners: "309956199498"
   ami_filter_name: 
   distribution: "RHEL"
   distribution_version: "8.2"
-  source_ami: "ami-0f6fa84b9ae33698d"
+  source_ami: ""
   ssh_username: "ec2-user"
   root_device_name: "/dev/sda1"
   volume_size: "15"
--- a/images/ami/rhel-82.yaml
+++ b/images/ami/rhel-82.yaml
@@ -3,7 +3,6 @@ download_images: true
 packer:
   ami_filter_name: "RHEL-8.2.0*2020*-x86_64-*"
   ami_filter_owners: "309956199498"
-  ami_filter_name: 
   distribution: "RHEL"
   distribution_version: "8.2"
   source_ami: ""
--- a/images/ami/rhel-82.yaml
+++ b/images/ami/rhel-82.yaml
@@ -9,7 +9,6 @@ packer:
   ssh_username: "ec2-user"
   root_device_name: "/dev/sda1"
   volume_size: "15"
-  remote_folder: "/home/ec2-user"
 
 build_name: "rhel-8.2"
 packer_builder_type: "amazon"
--- a/images/ami/rhel-82.yaml
+++ b/images/ami/rhel-82.yaml
@@ -1,14 +1,16 @@
 download_images: true
 
 packer:
-  ami_filter_name: "RHEL-8.2.0*2020*-x86_64-*"
-  ami_filter_owners: "309956199498"
+  # ami_filter_name: "RHEL-8.2.0*2020*-x86_64-*"
+  # ami_filter_owners: "309956199498"
+  ami_filter_name: 
   distribution: "RHEL"
   distribution_version: "8.2"
-  source_ami: ""
+  source_ami: "ami-0f6fa84b9ae33698d"
   ssh_username: "ec2-user"
   root_device_name: "/dev/sda1"
   volume_size: "15"
+  remote_folder: "/home/ec2-user"
 
 build_name: "rhel-8.2"
 packer_builder_type: "amazon"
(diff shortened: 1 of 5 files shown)


pkg/packer/manifests/aws/packer.json.tmpl
```diff
@@ -204,14 +210,15 @@
     },
     {
       "type": "shell",
+      "remote_folder":"{{user `remote_folder`}}",
       "inline": [
-        "mkdir -p /tmp/.goss-dir"
+        "mkdir -p ~/.goss-dir"
```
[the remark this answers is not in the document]
supershal: I think its a good idea. 
The default for `remote_folder` is empty which let packer to set it as `/tmp` for the `script` provisioners. 
We can explicitly default `remote_folder` to `/tmp` in the `variables` section and let user override it if they need to. that way we will be backward compatible too. @mkarroqe wdyt?

images/ami/rhel-82.yaml
```diff
@@ -3,12 +3,14 @@ download_images: true
...
   ami_filter_name: "RHEL-8.2.0*2020*-x86_64-*"
   ami_filter_owners: "309956199498"
+  ami_filter_name: 
   distribution: "RHEL"
   distribution_version: "8.2"
   source_ami: ""
   ssh_username: "ec2-user"
   root_device_name: "/dev/sda1"
   volume_size: "15"
+  remote_folder: "/home/ec2-user"
```
dkoshkin, line 13: Is this change needed? What about all the other images?

pkg/packer/manifests/aws/packer.json.tmpl
```diff
@@ -194,8 +201,7 @@
       "playbook_file": "./ansible/provision.yaml",
       "user": "{{user `ssh_username`}}",
       "ansible_env_vars": [
-        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes -o HostkeyAlgorithms=+ssh-rsa -o PubkeyAcceptedAlgorithms=+ssh-rsa'",
-        "ANSIBLE_REMOTE_TEMP='/tmp/.ansible/'"
+        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes -o HostkeyAlgorithms=+ssh-rsa -o PubkeyAcceptedAlgorithms=+ssh-rsa'"
```
dkoshkin, line 204: Similar to https://github.com/mesosphere/konvoy-image-builder/pull/559/files#r1010945066, to preserve existing behavior we would want something like:
```suggestion
        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes -o HostkeyAlgorithms=+ssh-rsa -o PubkeyAcceptedAlgorithms=+ssh-rsa'"
        "ANSIBLE_REMOTE_TEMP='{{ remote_folder }}/.ansible/'"
```

pkg/packer/manifests/aws/packer.json.tmpl
```diff
@@ -26,6 +26,7 @@
     "kubernetes_full_version": "",
     "manifest_output": "manifest.json",
     "python_path": "",
+    "remote_folder": "",
```
dkoshkin, line 29: I think here we want to default to Hashicrops default https://developer.hashicorp.com/packer/docs/provisioners/shell#remote_folder
```suggestion
    "remote_folder": "/tmp",
```

images/ami/rhel-82.yaml
supershal: we decided to keep this change only for RHEL-8.2 yaml to make sure at least one e2e test covers it. we didnt made this change to all other images since this parameter is optional. 
Do you have any preference on removing this from here or making the same change for all the AMIs?

pkg/packer/manifests/aws/packer.json.tmpl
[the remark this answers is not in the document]
supershal: made suggested change.

pkg/packer/manifests/aws/packer.json.tmpl
supershal: fixed.

pkg/packer/manifests/aws/packer.json.tmpl
supershal: fixed

images/ami/rhel-82.yaml
dkoshkin: Since these are visible to end users can we move this config to our RHEL 8.2 e2e config in TC instead?

dkoshkin (approved):
LGTM 🎉

alejandroEsc (approved):
nice!

images/ami/rhel-82.yaml
supershal: That works too. we have to test if KIB merges `packer:` block properly when provided as overrides. I will give it a try and move to TC config if works.

supershal: Tested and changed TC configuration of `E2E RHEL 8.2 Offline FIPS` build

supershal:
@mesosphere-ci help

PR #565, opened 2022-11-03
ci: regenerate release notes when replacing release artifacts

supershal:
- regenerate release notes when replacing the release artifacts
The [github action](https://github.com/mesosphere/konvoy-image-builder/blob/main/.github/workflows/release-tag.yaml#L138) for release replaces current released artifact with notarized artifacts, which deletes existing release notes
This fix should regenerate the release notes when replacing artifacts


supershal:
Closing this PR: The change suggested in the PR does not work. 
We need more testing on a fork of KIB and confirm if a fix works.

PR #592, opened 2022-11-29
fix: Enable fs.may_detach_mounts kernel parameter on RHEL 7.9, and variants

dlipovetsky:
Workaround for issue that affects the kernel used by RHEL 7.9, CentOS 7.9, and Oracle Linux 7.9. For issue, see https://bugzilla.redhat.com/show_bug.cgi?id=1441737.

* https://d2iq.atlassian.net/browse/D2IQ-94574

```release-note

```

### DIFF
--- a/ansible/roles/containerd/tasks/install.yaml
+++ b/ansible/roles/containerd/tasks/install.yaml
@@ -44,3 +44,26 @@
   loop:
     - ctr
     - containerd
+
+# Workaround for issue that affects the kernel used by RHEL 7.9, CentOS 7.9, and Oracle Linux 7.9
+# For issue, see https://bugzilla.redhat.com/show_bug.cgi?id=1441737.
+# Solution adapted from https://github.com/kubernetes-sigs/kubespray/blob/999586a110eed72914358a1b5bd182df514e755b/roles/kubernetes/preinstall/tasks/0080-system-configurations.yml#L90-L106.
+- name: Check if we need to set fs.may_detach_mounts
+  stat:
+    # The parameter is present only on affected kernels.
+    path: /proc/sys/fs/may_detach_mounts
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):
Nice!

ansible/roles/containerd/tasks/install.yaml
```diff
@@ -44,3 +44,26 @@
...
+# Solution adapted from https://github.com/kubernetes-sigs/kubespray/blob/999586a110eed72914358a1b5bd182df514e755b/roles/kubernetes/preinstall/tasks/0080-system-configurations.yml#L90-L106.
+- name: Check if we need to set fs.may_detach_mounts
+  stat:
+    # The parameter is present only on affected kernels.
+    path: /proc/sys/fs/may_detach_mounts
+    get_attributes: no
+    get_checksum: no
+    get_mime: no
+  register: fs_may_detach_mounts
+  ignore_errors: true
```
faiq, line 59: if we know the OS family + major distribution redhat + 7... lets add a when to check for this.

faiq (changes requested):
I would prefer to be more specific with detecting the sysctl setting

ansible/roles/containerd/tasks/install.yaml
dlipovetsky: My first instinct as well. However, this is really about the kernel, not the operating system version. Any kernel that has this parameter needs to have it enabled. In practice, the only kernel to have the parameter is the 3.10.x kernel used by RHEL 7.9, CentOS 7.9, and Oracle Linux 7.9.

faiq: That makes sense to me! Thank you for replying

PR #593, opened 2022-11-29
fix: propogate signals to packer sub-command

faiq:
propogates signal so we can safely clean up cloud resources

```release-note

```

### DIFF
--- a/pkg/packer/packer.go
+++ b/pkg/packer/packer.go
@@ -3,7 +3,10 @@ package packer
 import (
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"os/signal"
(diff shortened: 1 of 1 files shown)


dkoshkin (commented):
Oh yeah I remember this being an in issue in [Konvoy](https://github.com/mesosphere/konvoy/blob/8d174ec0634d2d21728ea457e511de2adc57fa6c/pkg/provisioner/terraform/command.go#L150-L161), but I don't remember why this was needed. Wha is the issue you were seeing that prompted this?

dkoshkin (approved):

supershal (commented):
Thanks for fixing this.

PR #596, opened 2022-12-02
e2e: uses magefile for e2e tests

faiq:
Should help clarify our Makefile mess 

```release-note

```


.github/workflows/aws-e2e.yaml
```diff
@@ -0,0 +1,105 @@
...
+  pull_request:
+    branches:
+    - '*'
+permissions:
+  contents: read
+  id-token: write
+env:
+  PACKER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+name: AWS
+jobs:
```
supershal, line 14: We also need to invoke these jobs on-demand from GHA UI using selecting "branch name", "build config" and "OS".
What do you think of using reusable-workflows. 
you can create another workflow with `workflow_call` and `workflow_dispatch` and call that work from current workflow which executes on PR. 

we did this in konvoy2.
caller workflow: https://github.com/mesosphere/konvoy2/blob/main/.github/workflows/daily_e2e_tests.yaml#L50-L54
which calls:
https://github.com/mesosphere/konvoy2/blob/main/.github/workflows/aws_e2e_tests.yaml#L10-L42

supershal: btw, if we decide to use reusable workflows, we can add this enhancement in follow up PR. This PR does not need to get complicated now.

alejandroEsc (commented):
very cool, a few questions and nits, nothing serious but awesome view of things to come

magefile.go
```diff
@@ -0,0 +1,464 @@
...
+		}
+	}
+	if !found {
+		return fmt.Errorf("buildInfra %s is invalid must be one of %v", buildInfra, validInfra)
+	}
+	return nil
+}
+
+func getBuildPath(buildOS, buildInfra string) string {
+	formattedOS := strings.TrimLeft(buildOS, " ")
```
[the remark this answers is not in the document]
faiq: we can but i think its better to be safe here

magefile.go
```diff
@@ -0,0 +1,464 @@
...
+var (
+	basic         = "basic"
+	fips          = "fips"
+	nvidia        = "nvidia"
+	offline       = "offline"
+	offlineFIPS   = "offline-fips"
+	offlineNvidia = "offline-nvidia"
+
+	validOS = []string{
+		"centos 7.9",
```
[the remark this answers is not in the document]
faiq: i guess we might as well! I'll refactor after this gets merged

.github/workflows/aws-e2e.yaml
```diff
@@ -0,0 +1,121 @@
+'on':
+  push:
+    branches:
+    - main
+  pull_request:
```
supershal, line 5: wdyt, if we just run it on pull_request until we completely migrate to TC?
and disable all TC configurations for e2e tests. just Keep TC configurations that runs on main and release which actually builds AMI without `--dry-run`.

.github/workflows/aws-e2e.yaml
```diff
@@ -0,0 +1,121 @@
...
+    branches:
+    - '*'
+permissions:
+  contents: read
+  id-token: write
+name: AWS
```
supershal, line 11:
```suggestion
name: E2E  Tests - Build AWS AMI
```

.github/workflows/aws-e2e.yaml
```diff
@@ -0,0 +1,121 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.18.0'
```
supershal, line 97: should get go version from go.mod file. https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

.github/workflows/aws-e2e.yaml
faiq: changed to just pull requests and will do the runs without `--dry-run` when refactoring to a shared workflow

.github/workflows/aws-e2e.yaml
```diff
@@ -0,0 +1,118 @@
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/konvoy-image-builder
+          aws-region: us-west-2
+
+      - name: Run Mage
+        uses: magefile/mage-action@v2
+        with:
+          version: latest
+          args: runE2e "${{ matrix.os }}" "${{ matrix.buildConfig }}" aws true
```
supershal, line 116: it would be cool if we could use cmd flags.
`rune2e --os=rhel-8.2 --buildconfig=offline --provider aws --dry-run=true`
https://github.com/magefile/mage/pull/328

Dockerfile.devkit
```diff
@@ -135,9 +135,8 @@ RUN if [ "${DOCKER_GID}" -ne "${GROUP_ID}" ]; then \
 # place it in /usr/share/ansible/collections, the container will be run with a different user
 RUN mkdir -p /usr/share/ansible/collections && ansible-galaxy collection install ansible.utils -p /usr/share/ansible/collections
 
-
 # hadolint ignore=DL3059
-RUN packer plugins install github.com/hashicorp/googlecompute ">=1.0.0" \
+RUN --mount=type=secret,id=githubtoken PACKER_GITHUB_API_TOKEN=$(cat /run/secrets/githubtoken) packer plugins install github.com/hashicorp/googlecompute ">=1.0.0" \
```
supershal, line 139: where is `/run/secrets/githubtoken` created?

supershal (approved):
LGTM. Good first iteration. We will need couple of more iterations to make it good enough.

Dockerfile.devkit
faiq: it gets created by the secret mount when we pass in `--secret ` in our docker buildx build command

PR #598, opened 2022-12-07
fix: dont ipv6 forward if you dont have ipv6 forwarding file 

faiq:
checks before enabling ipv6 forwarding if ipv6 is present

* https://d2iq.atlassian.net/browse/D2IQ-94653

```release-note
checks if ipv6 is enabled to set forwarding
```

### DIFF
--- a/ansible/roles/networking/tasks/main.yaml
+++ b/ansible/roles/networking/tasks/main.yaml
@@ -28,6 +28,22 @@
     state: present
     reload: yes
 
+- name: stat ipv6 in proc/sys/net
+  stat:
+    path: /proc/sys/net/ipv6
+  register: ipv6
+
+- name: "check if ipv6 is disabled"
+  lineinfile:
+    dest: /proc/sys/net/ipv6/conf/default/disable_ipv6
+    line: "1"
+    state: present
+  become: yes
+  check_mode: yes
+  register: ipv6Disabled
+  when:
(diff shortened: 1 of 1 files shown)


ansible/roles/networking/tasks/main.yaml
```diff
@@ -28,13 +28,32 @@
...
     reload: yes
 
+- name: stat ipv6 in proc/sys/net
+  stat:
+    path: /proc/sys/net/ipv6
+  register: ipv6
+
+- name: "check if ipv6 is disabled"
+  lineinfile:
+    dest: /proc/sys/net/ipv6/conf/default/disable_ipv6
```
dkoshkin, line 38: I did some googling, and there are other ways to disable https://techdocs.broadcom.com/us/en/ca-enterprise-software/it-operations-management/network-flow-analysis/21-2/installing/system-recommendations-and-requirements/linux-servers/disable-ipv6-networking-on-linux-servers.html/ (and maybe others?), where I think 
`/proc/sys/net/ipv6/conf/default/disable_ipv6` will not exist.


From https://d2iq.atlassian.net/browse/D2IQ-94653
>sysctl: cannot stat /proc/sys/net/ipv6/conf/all/forwarding:
Maybe we can just check if file exists `/proc/sys/net/ipv6/conf/all/forwarding` instead?

dkoshkin (approved):

PR #627, opened 2023-01-13
feat: support building Ubuntu 20.04 with offline OS packages bundle

dkoshkin:
Changes to support creating Ubuntu 20.04 images using offline OS packages bundle.

`apt` doesn't have similar functionality to `yum` to `--disablerepo/enablerepo`, instead when using the offline bundles we disable the other repos and then re-enable it back in sysprep.

* https://d2iq.atlassian.net/browse/D2IQ-94124

```release-note

```


dkoshkin:
Tested locally using an air-gapped AWS AMI.
This should translate to vSphere as there are no additional packages that are needed there.

ansible/roles/sysprep/tasks/debian.yml
```diff
@@ -32,16 +32,14 @@
       - /etc/apt/sources.list.d
     patterns: '*.list.disabled'
   register: repo_files
-  when:
-    disable_public_repos | default(false) | bool
-    and reenable_public_repos | default(true) | bool
+  when: (disable_public_repos | default(false) | bool and reenable_public_repos | default(true) | bool) or
```
supershal, line 35: 😺 we should revisit to simplify this in next PR.

dkoshkin: tbh these vars are not used anywhere else and are not documented. We could remove it in the future but didn't want to do that here.

dlipovetsky (approved):

supershal (approved):

Issue #464, opened 2022-08-11
Adding user tags is impossible


mimmus:
Our organization has some mandatory tag for EC2 instances, creating also a temporary instance without tags is impossible.
Looking at Packer template, adding custom tags is actually impossible.

faiq:
Adding user tags is not impossible, you can make customizations to the packer manifest that kib produces and use it for your builds.

For example if I need to create an image with with custom tags 

```json 
{
"my_cool": "tag"
}
``` 

I would do it by running the following steps

1. Generate manifest files for the KIB run by doing 

```
$ ./konvoy-image build images/ami/centos-79.yaml  
e2022/08/11 19:01:17 writing new packer configuration to work/centos-7-1660244477-JpGye
2022/08/11 19:01:17 starting packer build
centos-7.9: output will be in this color.
``` 

and sending `SIGINT` to the process to kill it 

2. Next edit the file `packer.json` from the above work directory. In my case the path was `work/centos-7-1660244477-JpGye/packer.json` 

3. When editing go to builders section of the json and add a [run_tags](https://www.packer.io/plugins/builders/amazon/ebs#run_tags) block. When looking at it with jq i get the following result: 

```
$ cat work/centos-7-1660244477-JpGye/packer.json | jq '.builders|.[]|."run_tags"'
{
  "my_cool": "tag"
}
``` 

4. Pass the modified packer file to konvoy-image like so:

`$ ./konvoy-image build --packer-manifest=work/centos-7-1660244477-JpGye/packer.json images/ami/centos-79.yaml` 

You will see the logs 

```
    centos-7.9: Found Image ID: ami-0686851c4e7b1a8e1
==> centos-7.9: Creating temporary keypair: packer_62f554df-e60d-8919-4b0e-3c9ba405551c
==> centos-7.9: Creating temporary security group for this instance: packer_62f554e6-3004-1d78-1454-fa28bb3da4e7
==> centos-7.9: Authorizing access to port 22 from [0.0.0.0/0] in the temporary security groups...
==> centos-7.9: Launching a source AWS instance...
    centos-7.9: Adding tag: "my_cool": "tag"
    centos-7.9: Instance ID: i-0be42f1c3f6455ad7
``` 

And the instance will be created with the tag you need. 


We will be adding a feature to pass this in via command line, but until then this is how you can do it.

mimmus:
This workaround is really useful.
Thanks

faiq:
[...]

PR #630, opened 2023-01-19
feat: allow submaps in packer object

faiq:
#464 allows submaps in the packer subsection

* https://d2iq.atlassian.net/browse/D2IQ-92895

```release-note
feat: allows users to put maps in packer config 
```

### DIFF
--- a/pkg/app/config.go
+++ b/pkg/app/config.go
@@ -295,6 +295,19 @@ func GenPackerVars(config Config, extraVarsPath string) ([]byte, error) {
 			case []string:
 				p[key] = strings.Join(v, ",")
 			default:
+				if reflect.ValueOf(v).Kind() == reflect.Map {
+					m := make(map[string]interface{})
+					reflectOfV := reflect.ValueOf(v)
+					for _, e := range reflectOfV.MapKeys() {
+						if subKey, ok := e.Interface().(string); ok {
+							subVal := reflect.ValueOf(v).MapIndex(e).Interface()
+							m[subKey] = subVal
+						}
+					}
+					p[key] = m
+					break
+
+				}
 				p[key] = fmt.Sprintf("%v", v)
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/app/config.go
```diff
@@ -295,6 +295,18 @@ func GenPackerVars(config Config, extraVarsPath string) ([]byte, error) {
 			case []string:
 				p[key] = strings.Join(v, ",")
 			default:
+				reflectOfV := reflect.ValueOf(v)
+				if reflectOfV.Kind() == reflect.Map {
+					m := make(map[string]interface{})
```
supershal, line 300: why further conversion to `map[string]interface{}` needed? `p[key] = v` instead works as expected. Do we need to explicitly check if keys of the `run_tags` are string?

I tested with following
```go
  
default:
	reflectOfV := reflect.ValueOf(v)
	switch reflectOfV.Kind() {
	case reflect.Map:
		p[key] = v
	default:
		p[key] = fmt.Sprintf("%v", v)                           
```

faiq: i think it would be make most sense to import a map of stings to values. it seems a bit more safe to be explicit when working with untyped objects

pkg/app/config.go
```diff
@@ -295,6 +295,18 @@ func GenPackerVars(config Config, extraVarsPath string) ([]byte, error) {
...
+					for _, e := range reflectOfV.MapKeys() {
+						if subKey, ok := e.Interface().(string); ok {
+							subVal := reflectOfV.MapIndex(e).Interface()
+							m[subKey] = subVal
```
supershal, line 304: nit
```suggestion
							m[subKey] = reflectOfV.MapIndex(e).Interface()
```

supershal (approved):
🎉

pkg/app/config.go
dlipovetsky: I also found it hard to understand the need for reflect.

I guess the input (`v`) can only have a depth of two (keys, and subkeys), and when we copy it over, we want to skip any subkeys that aren't strings?

I guess this is because something like Shalin's suggestion, or this

```go
			case map[string]interface{}:
				p[key] = v
			default:
				p[key] = fmt.Sprintf("%v", v)
```

would panic if the subkey is a not a string, e.g., it is an integer.

PR #631, opened 2023-01-20
feat: Install specific cri-tools package on debian derivatives

dlipovetsky:
The cri-tools package is a dependency of kubeadm. However, kubeadm installs the latest cri-tools version. And while the kubeadm is compatible with the container runtime, the latest cri-tools version may not be.

For example, on January 20, 2023, kubeadm v1.25.x installs cri-tools v1.26.0. The former is compatible with containerd v1.5.x, but the latter is not.

Please see https://github.com/kubernetes/release/issues/2866 for more details.

* https://d2iq.atlassian.net/browse/D2IQ-95429

```release-note
On Debian-based Linux distributions, install a version of the cri-tools package known to be compatible with both the Kubernetes and container runtime versions.
```

### DIFF
--- a/ansible/group_vars/all/defaults.yaml
+++ b/ansible/group_vars/all/defaults.yaml
@@ -3,12 +3,27 @@ python_path: ""
 
 # This is also in images/common.yaml as that's where the go code expects it to be.
 # If it's not there, the kubernetes_full_version will have "None" for a version number.
+#
+# IMPORTANT When you update kubernetes_version, also update crictl_version.
 kubernetes_version: "1.25.4"
-
-containerd_version: "1.6.10"
-kubernetes_cni_version: "0.9.1"
-crictl_version: "1.22.0"
 kubernetes_semver: "v{{ kubernetes_version }}"
+
+kubernetes_cni_version: "0.9.1"
+
+# The crictl CLI is released as part of the http://sigs.k8s.io/cri-tools project.
+# The project release closely follows the Kubernetes release cycle, and uses a
+# nearly identical version scheme.
+# IMPORTANT When you update crictl_version, also update crictl_sha256.
+crictl_version: "1.22.0"
+
+# On flatcar Linux, we install crictl from a release artifact, not a system package.
+# The url points to the linux/amd64 release artifact.
+crictl_url: https://github.com/kubernetes-sigs/cri-tools/releases/download/v{{ crictl_version }}/crictl-v{{ crictl_version }}-linux-amd64.tar.gz
+# The sha256 sum verifies the integrity of the release artifact.
+crictl_sha256: 45e0556c42616af60ebe93bf4691056338b3ea0001c0201a6a8ff8b1dbc0652a
+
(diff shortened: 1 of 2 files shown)


ansible/roles/kubeadm/tasks/debian.yaml
```diff
@@ -2,6 +2,27 @@
...
+    apt-get install --force-yes --yes \
+      cri-tools={{ critools_deb }}
+  args:
+    warn: false
+  register: result
+  until: result is success
+  retries: 3
+  delay: 3
+
+- name: add version hold for cri-tools package
```
dkoshkin, line 23: Great idea!

ansible/group_vars/all/defaults.yaml
```diff
@@ -43,7 +61,6 @@ kubernetes_cni_semver: v{{ kubernetes_cni_version }}
 kubernetes_cni_http_checksum: sha256:https://storage.googleapis.com/k8s-artifacts-cni/release/{{ kubernetes_cni_semver }}/cni-plugins-linux-amd64-{{ kubernetes_cni_semver }}.tgz.sha256
 kubernetes_cni_http_source: https://github.com/containernetworking/plugins/releases/download
 crictl_url: https://github.com/kubernetes-sigs/cri-tools/releases/download/v{{ crictl_version }}/crictl-v{{ crictl_version }}-linux-amd64.tar.gz
-crictl_sha256: 45e0556c42616af60ebe93bf4691056338b3ea0001c0201a6a8ff8b1dbc0652a
```
dkoshkin, line 46: nit: `crictl_sha256` used with `crictl_url` for Flatcar, can we keep these vars together please.

dlipovetsky: I moved `crictl_sha256` because it must change whenever `crictl_version` changes. I'll think about this some more.

ansible/roles/kubeadm/tasks/debian.yaml
dlipovetsky: Do you have any thoughts about how the two kubeadm and cri-tools holds interact? I'm a little worried the order matters, but [man page](https://manpages.ubuntu.com/manpages/bionic/man8/apt-mark.8.html) isn't enlightening.

ansible/group_vars/all/defaults.yaml
dlipovetsky: Fixed

cneth (approved):

dlipovetsky:
Once this merges, I'll backport to `release-1.24` (DKP 2.4.x) and `release-1.19` (DKP 2.3.x)

ansible/roles/kubeadm/tasks/debian.yaml
dlipovetsky: Good point, thanks! I added a comment explaining the hold.

dlipovetsky:
Added a comment, squashed it into an earier commit, and force-pushed.

PR #637, opened 2023-01-24
feat: build rocky linux 9.1 images

supershal:
- E2E tests for rocky linux 9.1

* https://d2iq.atlassian.net/browse/D2IQ-94873

We had disabled yum version lock for RHEL 8 because of dependency errors with already installed packets.
https://github.com/mesosphere/konvoy-image-builder/blob/main/ansible/roles/setup_versionlock/tasks/redhat.yaml#L2-L5
We had fixed this issue by not upgrading already installed packages.
We will need to revisit this logic for RHEL 8 and see if we can add back this version lock for RHEL 8.
I will test and file separate PR for this.

```release-note

```


ansible/roles/containerd/tasks/redhat.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: remove versionlock for containerd
-  command: yum versionlock delete containerd.io
+  command: yum versionlock {{ '--disablerepo "*" --enablerepo offline' if offline_mode_enabled else '' }} delete containerd.io
```
supershal, line 3: yum versionlock commands try to download all mirrors from internet.  Adding logic for preventing yum from updating mirrors when using `yum versionlock` in air-gapped installations for "RHEL" ansible distributions.

make/ci.mk
```diff
@@ -79,6 +80,10 @@ e2e.build.rhel-8.6: rhel86
 
 e2e.build.rhel-8.6-offline: rhel86-offline infra.aws.destroy
 
+e2e.build.rocky-9.1-offline: rocky91-offline infra.aws.destroy
+
+e2e.build.rocky-9.1: rocky91
+
 e2e.build.rhel-8.6-nvidia: rhel86-nvidia
```
faiq, line 88: should we not do these any more?

ansible/group_vars/all/defaults.yaml
```diff
@@ -31,7 +31,7 @@ critools_deb: "{{ crictl_version }}-00"
 critools_rpm: "{{ crictl_version }}-0"
 
 # Containerd container runtime version.
-containerd_version: "1.6.10"
+containerd_version: "1.6.15"
```
faiq, line 34: Any reason in particular we need to bump this as a part of this PR?

supershal: Thanks for bringing this up.
This is the latest and only containerd `1.6.x` version we are building for rocky. https://teamcity.mesosphere.io/buildConfiguration/MesosphereOnly_ClosedSource_DkpReleases_UpstreamBuilds_Containerd_Rocky91?branch=&buildTypeTab=overview&mode=builds

I agree this will impact other OS builds too. Do you see bumping patch version of containerd here would require any additional testing and considerations?

I can manually build containerd `1.6.10` for rocky too and revert this change.

make/ci.mk
supershal: I am keeping it in sync with other related targets for other OS. We can remove all related target at once when we deprecate them in separate PR.

ansible/group_vars/all/defaults.yaml
faiq: I think it should be fine to do as a part of this PR. I was just wondering why you did it.

dkoshkin: I think we do want to bump the version but I would really prefer it as a separate PR, and then base this PR on that. WDYT?

supershal: Thats fair. I will revert this commit and file separate PR

images/azure/rocky-91.yaml
```diff
@@ -0,0 +1,13 @@
+---
+download_images: true
+
+packer:
+  distribution: "rockylinux-9"                # Offer
+  distribution_version: "rockylinux-9"  # SKU
+  image_publisher: "erockyenterprisesoftwarefoundationinc1653071250513"
```
supershal, line 7: This is ID for the official free rocky linux image template. https://portal.azure.com/#view/Microsoft_Azure_Marketplace/GalleryItemDetailsBladeNopdl/id/erockyenterprisesoftwarefoundationinc1653071250513.rockylinux-9

ansible/group_vars/all/defaults.yaml
supershal: reverted back to containerd 1.6.10

supershal:
Created TC configurations for building Rockylinux release AMI (fips, non-fips) , Azure image (non-fips)

ansible/roles/containerd/tasks/redhat.yaml
dkoshkin: Wow so that means the rockylinux already comes wit Containerd installed? 🤔

ansible/roles/kubeadm/tasks/redhat.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: remove versionlock for kubeadm packages
-  command: yum versionlock delete kubeadm
+  command: yum versionlock {{ '--disablerepo "*" --enablerepo offline' if offline_mode_enabled else '' }} delete kubeadm
```
dkoshkin, line 3: How are we not seeing these issues for RHEL and CentOS, does `yum versionlock` not try to download the metadata for those?

images/azure/rocky-91.yaml
dkoshkin: 🤮

make/images.mk
```diff
@@ -443,6 +443,23 @@ rhel86-fips-offline:
...
+# Rocky 9.1
+.PHONY: rocky91
+rocky91:
+	$(MAKE) build-aws-rocky-9.1
+
+.PHONY: rocky91-offline
+rocky91-offline:
+	$(MAKE) aws-rocky-9.1_offline
+
+.PHONY: rocky91-fips-offline
```
dkoshkin, line 455: Did we build FIPS offline packages? I do't think RockyLinux supports the correct `140-2` FIPS version only https://rockylinux.org/news/certifications-fips-2022-06-11/ `140-3`.

images/azure/rocky-91.yaml
```diff
@@ -4,9 +4,13 @@ download_images: true
 packer:
   distribution: "rockylinux-9"                # Offer
   distribution_version: "rockylinux-9"  # SKU
+  # Azure Rocky linux official image: https://portal.azure.com/#view/Microsoft_Azure_Marketplace/GalleryItemDetailsBladeNopdl/id/erockyenterprisesoftwarefoundationinc1653071250513.rockylinux-9
   image_publisher: "erockyenterprisesoftwarefoundationinc1653071250513"
   image_version: "latest"
   ssh_username: "azureuser"
+  plan_image_sku: "rockylinux-9" # SKU
```
supershal, line 11: Azure plan information needed to use marketplace image. 
Packer throwing following error when plan information was not provided. 
`Creating a virtual machine from Marketplace image or a custom image sourced from a Marketplace image requires Plan information in the request`

make/images.mk
supershal: FIPS offline packages for rocky. [Link](https://teamcity.mesosphere.io/buildConfiguration/MesosphereOnly_ClosedSource_DkpReleases_OsPackagesBundles_Rocky9fips/3789491?buildTab=log&expandAll=true&focusLine=2019)
I have still not run any FIPS validation tests

images/azure/rocky-91.yaml
dkoshkin: This thread has lots of useful information https://forums.rockylinux.org/t/azure-rocky-image-on-marketplace/5230/46

ansible/roles/containerd/tasks/redhat.yaml
dkoshkin: Discussed offline, the behavior for RockyLinux is different from CentoOS/RHEL `yum` and it tries to reach other repos when running `yum versionlock`.
But this is a good generic change that will prevent this bug in future RHEL releases too.

dkoshkin (approved):

images/azure/rocky-91.yaml
supershal: Thank you for looking it up

faiq (approved):

PR #638, opened 2023-01-24
ci: change vsphere e2e test's datastore to ovh-netapp-users

supershal:
- vsphere datastore `ovh-nfs` is deprecated. moving the template creation to `ovh-netapp-users`


dkoshkin (approved):
Thanks!

PR #648, opened 2023-02-01
ci: adds azure GHA test

faiq:
Moves azure tests to GHA 

```release-note

```


Makefile
```diff
@@ -222,7 +222,7 @@ github-token.txt:
 .PHONY: buildx
 buildx:
 buildx:
-	 docker buildx create --driver=docker-container --use --name=konvoy-image-builder || true
+	 docker buildx create --use --name=konvoy-image-builder || true
```
dkoshkin, line 225: Why was this change needed?

.github/workflows/azure-e2e.yaml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  rune2e:
+    strategy:
+      fail-fast: false
+      max-parallel: 10
+      matrix:
+        os:
+        - "ubuntu 20.04"
+        buildConfig:
+        - "basic"
```
dkoshkin, line 18: Im loving how easy it is to add new configs to the build matrix!

Makefile
faiq: It's not needed, but it makes things a lot faster because the cache can be re-used

faiq: I can pull it out to separate PR if you'd like

.github/workflows/azure-e2e.yaml
```diff
@@ -0,0 +1,52 @@
+'on':
+  pull_request:
+    branches:
+    - '*'
```
supershal, line 4: I think this filter is not needed.

magefile.go
```diff
@@ -230,8 +233,10 @@ func getBuildPath(buildOS, buildInfra string) string {
 	formattedOS = strings.Replace(formattedOS, ".", "", 1)
 
 	formattedOS = strings.Replace(formattedOS, "redhat", "rhel", 1)
-	formattedOS = strings.Replace(formattedOS, "1804", "18", 1)
-	formattedOS = strings.Replace(formattedOS, "2004", "20", 1)
+	if buildInfra == aws {
+		formattedOS = strings.Replace(formattedOS, "1804", "18", 1)
+		formattedOS = strings.Replace(formattedOS, "2004", "20", 1)
+	}
```
faiq, line 239: This is annoying we have different file names for azure and aws

.github/workflows/azure-e2e.yaml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
```
supershal, line 30: please use `cache: true` to run this action faster

.github/workflows/azure-e2e.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Docker Login
+        uses: docker/login-action@v2
+        with:
+          registry: docker.io
+          username: mesosphereci
```
supershal, line 36: can we use org level secrets. `DOCKER_READ_WRITE_USERNAME` and `DOCKER_READ_WRITE_PASSWORD` ?
Assuming that they should be ok to use even it for a public repo.

Makefile
dkoshkin: Yes, that would be good.

supershal: This was added when we were trying arm64 support. We later find out that we dont need it. This argument `--driver=docker-container` unnecessary slows builds down and does not cache docker file system layers between builds.

.github/workflows/azure-e2e.yaml
faiq: Those are not org level secrets

magefile.go
supershal: 💯

Makefile
faiq: removed and will file another PR

magefile.go
faiq: i changed the file names as a part of this PR

.github/workflows/azure-e2e.yaml
faiq: Used other secrets

supershal (approved):
lgtm

dkoshkin (approved):
Lets wait for the new Azure and AWS tests to pass before mergning, but this is great!

supershal:
Test looks good. marked Ubuntu tests GHA as required and removed same tests in TC from the `required` list. The TC jobs are paused as well.

PR #649, opened 2023-02-01
feat: build vsphere template for rockylinux 9.1

supershal:
- build rocky linux 9.1 vsphere template

* https://d2iq.atlassian.net/browse/D2IQ-95242

The rockylinux base template is created using packer HCL template. https://github.com/mesosphere/konvoy-image-builder/compare/shalin/vsphere-rocky-base 
This work still needs some polishing. I will file separate PR for this later because this reference implementation is not blocking.

### DIFF
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -35,7 +35,7 @@
     packages:
       - cloud-init
       - cloud-utils-growpart
-      - python2-pip
+      - "python{{ '3' if ansible_distribution_major_version == '9' else '2'}}-pip"
   when: ansible_os_family == "RedHat"
 
 - name: Install cloud-init packages
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -35,7 +35,7 @@
     packages:
       - cloud-init
       - cloud-utils-growpart
-      - "python{{ '3' if ansible_distribution_major_version == '9' else '2'}}-pip"
+      - "python{{ '3' if ansible_distribution_major_version|int >= 9 else '2'}}-pip"
   when: ansible_os_family == "RedHat"
 
 - name: Install cloud-init packages
--- a/ansible/roles/sysprep/tasks/redhat.yml
+++ b/ansible/roles/sysprep/tasks/redhat.yml
@@ -41,10 +41,17 @@
 - name: Remove yum package lists
   command: "{{ sysusr_prefix }}/bin/yum -y clean all"
 
-- name: Reset network interface IDs for RHEL 8.x and below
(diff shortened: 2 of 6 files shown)


ansible/roles/sysprep/tasks/redhat.yml
```diff
@@ -41,8 +41,36 @@
...
+  with_items: "{{  network_profiles_sysconfig.files | map(attribute='path') | list }}"
+
+# NetworkManager stores new network profiles in keyfile format in the
+# /etc/NetworkManager/system-connections/ directory.
+
+# Previously, NetworkManager stored network profiles in ifcfg format
+# in this directory (/etc/sysconfig/network-scripts/). However, the ifcfg
+# format is deprecated. By default, NetworkManager no longer creates
+# new profiles in this format.
+- name: find network profiles at /etc/NetworkManager/system-connections
```
dkoshkin, line 63: Curious how did you know that this is needed?

dkoshkin (approved):

ansible/roles/sysprep/tasks/redhat.yml
supershal: I found this during testing offline mode. We do not update/install common rpms if it is already installed on the system. 
The network manager version that comes with RockyLinux DVD ISO uses new format and location for network profiles. New version of NetworkManager drops a readme(deprecation notice) at existing location `/etc/sysconfig/network-scripts/` explaining about new location and format.

It is possible that the rockylinux base AMI and azure uses a NetworkManager version that still uses `/etc/sysconfig/network-scripts/` for network profiles evident by previously passing aws and azure e2e tests. 
So I created these idempotent tasks to handle both case without guessing networkManger version or OS version.

ansible/roles/providers/tasks/vmware.yml
```diff
@@ -35,7 +35,7 @@
     packages:
       - cloud-init
       - cloud-utils-growpart
-      - python2-pip
+      - "python{{ '3' if ansible_distribution_major_version == '9' else '2'}}-pip"
```
faiq, line 38: So we only want python 3 for rhel or rocky 9?

faiq (approved):

ansible/roles/providers/tasks/vmware.yml
supershal: Rocky linux claims to be compatible with RHEL from version 9 and newer. rocky 9.x uses python3-* packages. We dont support RHEL 9.X now, but we will need python3-* for RHEL too. condition `ansible_distribution_major_version == '9'` will cover both usecaes. however, now I realize that it would be better set it for 9.x and above.

I can change the condition to `(ansible_distribution_major_version | int)  >= 9`.  wdyt ?

faiq: that makes sense to me. 

my comment was thinking ahead to the point where we do support rhel 9.

PR #657, opened 2023-02-06
ci: bump CAPPP on successful release jobs

supershal:
Prevents CAPPP bump in case of KIB release failure

### DIFF
--- a/.github/workflows/release-kib.yaml
+++ b/.github/workflows/release-kib.yaml
@@ -44,9 +44,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.MESOSPHERECI_USER_TOKEN }}
 
   bump-kib:
-    needs: [release-artifacts, sign-binary]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


faiq (commented):
i dont think it needs sign binary here

supershal:
> i dont think it needs sign binary here

Agree. The CAPPP bump does not need signed binary job technically. However in my opinion, the KIB release should not considered successful if all steps, including signing binary, completes successfully. IMHO, bumping CAPPP needed only on successful release.

PS: slack notification needed for successful/failed release job. I will file separate PR for that.

dkoshkin (approved):

faiq:
We can rerun the sign binary step separately. The two things are not related so it should not block the release

dkoshkin:
@faiq so its possible to trigger the MacOS signing job if it fails without rerunning the rest of the job?
Would that be the workflow we want or would we instead want to rerun the whole release job?

faiq (approved):

PR #659, opened 2023-02-06
ci: clean up github actions workflows

supershal:
- use name for the actions, steps wherever it was missing
- use go version from `go.mod` file
- remove coverage.yaml workflow. merged it in the unit-tests.yaml workflow.
- merge golang-ci and lint workflow into one file
- run build, lint and test job on tags
- use d2iq docker mirror registry


dkoshkin (approved):
Thank you for the cleanup 🙏

.github/workflows/build.yml
```diff
@@ -6,22 +6,23 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v*'
```
faiq, line 10: Why do we want to run on tags?

.github/workflows/lint.yml
```diff
@@ -4,27 +4,29 @@ name: Lint
 # yamllint disable-line rule:truthy
 on:
   push:
+    tags:
+      - 'v*'
```
faiq, line 8: Again why do we need to run this on tags?

faiq (approved):
Some questions

.github/workflows/build.yml
supershal: My initial thought was that if for any reason we have to manually force push a release tag, we should run basic workflows build, test and lint.
The current `make release` target does not cover all of them.

faiq: Gotcha. makes sense

PR #661, opened 2023-02-07
ci: add rocky 9.1 builds configuration for azure

supershal:
- Add ubuntu 18.04 and rocky 9.1 e2e tests for azure.

```release-note

```

### DIFF
--- a/.github/workflows/azure-e2e.yaml
+++ b/.github/workflows/azure-e2e.yaml
@@ -13,7 +13,6 @@ jobs:
       matrix:
         os:
         - "ubuntu 20.04"
-        - "ubuntu 18.04"
         - "rocky 9.1"
         buildConfig:
         - "basic"
--- a/.github/workflows/azure-e2e.yaml
+++ b/.github/workflows/azure-e2e.yaml
@@ -13,6 +13,8 @@ jobs:
       matrix:
         os:
         - "ubuntu 20.04"
+        - "ubuntu 18.04"
+        - "rocky 9.1"
         buildConfig:
         - "basic"
     runs-on: self-hosted


supershal:
We also have azure tests for "rhel-7", "oracle-7", "sles-15" and "flatcar" in teamcity which are not migrated over GHA.
These configurations are not supported by https://docs.d2iq.com/dkp/2.4/supported-infrastructure-operating-systems

Not sure why those builds added in Teamcity. Should we migrate them to GHA too? 
or should we disable them in TC?
@faiq @dkoshkin

dkoshkin (approved):

.github/workflows/azure-e2e.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
       matrix:
         os:
         - "ubuntu 20.04"
+        - "ubuntu 18.04"
```
faiq, line 16:
```suggestion
```

faiq (changes requested):
we dont support ubuntu 18 on azure

faiq:
> We also have azure tests for "rhel-7", "oracle-7", "sles-15" and "flatcar" in teamcity which are not migrated over GHA.
These configurations are not supported by https://docs.d2iq.com/dkp/2.4/supported-infrastructure-operating-systems
> Not sure why those builds added in Teamcity. Should we migrate them to GHA too?
or should we disable them in TC?
@faiq @dkoshkin

we shouldn't add builds that we don't claim we support. adding them to the previous CI was a mistake

.github/workflows/azure-e2e.yaml
supershal: We support Ubuntu 18.04 for AKS. https://docs.d2iq.com/dkp/2.3/supported-infrastructure-operating-systems#id-(2.3)SupportedInfrastructureOperatingSystems-AKS 
However I notice that konvoy CLI does not use "gallery image" flag for aks cluster/nodepool create. 
so just to confirm, aks users will always use upstream azure image?

DreadPirateShawn:
> we shouldn't add builds that we don't claim we support

I'm inclined to agree.

.github/workflows/azure-e2e.yaml
dkoshkin: Correct DKP users will use the default upstream images for AKS (and EKS).
I agree with @faiq we should only add configs that DKP supports.

supershal: Thank you.

faiq (approved):

PR #674, opened 2023-02-09
ci: enable manual trigger for GHA build and e2e jobs

supershal:
- allow user to trigger GHA build manually.


dkoshkin (approved):
Thanks

PR #692, opened 2023-02-23
fix: set force = true and update k8s http source

faiq:
https://d2iq.atlassian.net/browse/D2IQ-95908

* https://d2iq.atlassian.net/browse/D2IQ-95908

```release-note
fix: forces all items that use get_url to be downloaded 
```


dkoshkin (approved):
Looks great!
nit: what do you think about squashing the last 3 commits to keep to history cleaner? We'll probably want to not squash merge to allow the first commit to be cherry-picked and backported.

faiq:
Yes, let me squash the first three!

supershal (approved):

PR #696, opened 2023-02-27
ci: add goss command spec

faiq:
adds a command to check kubeadm version 

```release-note

```

### DIFF
--- a/goss/flatcar/command.yaml
+++ b/goss/flatcar/command.yaml
@@ -1,0 +1,11 @@
+command:
+  version:
+    exit-status: 0
+    exec: "kubeadm version -o short"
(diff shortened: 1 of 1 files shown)


goss/flatcar/command.yaml
```diff
@@ -0,0 +1,11 @@
+command:
+  version:
+    exit-status: 0
+    exec: "kubeadm version -o short"
+    # update this when we change kubernetes version to `
+    stdout:
+      - v1.25.4
```
supershal, line 7: The kubernetes version is not strictly fixed in all usecases. We allow to override using CLI flag and override files too. 
is this any reason this is strictly needed for flatcar?

PR #698, opened 2023-02-28
ci: revert setting github token in devkit container for release-2.4

supershal:
Revert Dockerfile.devkit changes back to release-2.4 head.


faiq:
we were able to fix it using devkit

PR #701, opened 2023-03-01
fix: fallback to ansible_user_id when using root user

dkoshkin:
Fix from https://github.com/kubernetes-sigs/image-builder/pull/739/files#diff-2f3020c3867a75d1e2395d10238225e0fa343e8097cb434f647817e446354a88R149

Looks like when using `root` as the SSH user, `ansible_env.SUDO_USER` is not set.
Fallback to using `ansible_user_id` which will in this case already equal the ssh user.

```release-note

```


supershal (approved):
Thank you

PR #553, opened 2022-10-28
WIP establish regular molecule tests

fatz:
**What problem does this PR solve?**:
Currently only basic testing is done against a full image build. But changes in ansible are not covered by tests. This actually lead to customer facing issue.

Molecule was in a quite unmaintained state. To improve code quality its important to establish a test driven approach making changes to the ansible code in KIB.

This PR will establish automated molecule testing for all supported OSes by using AWS instances for testing.

We'll not cover provider specific code but focus on the supported Distributions

Action items:
[x] make sure ansible ( without the sysprep step ) is idempotent so while developing new features ansible could be run over and over again
[ ] ensure molecule works from within the devkit container
[ ] establish private AWS GH runner for this repository
[ ] add github actions to run the molecule suite whenever there are changes in `/ansible`
[ ] add github action to run special GPU related tests whenever the `/ansible/roles/gpu` has changes

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-NUMBER

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #713, opened 2023-03-02
fix: set limitnofile

faiq:
https://d2iq.atlassian.net/browse/D2IQ-96107
This change in contained seems to cause problems for workloads that use disk https://github.com/containerd/containerd/pull/4475

```release-note

```

### DIFF
--- a/ansible/roles/containerd/tasks/main.yaml
+++ b/ansible/roles/containerd/tasks/main.yaml
@@ -4,6 +4,12 @@
     path: "/etc/systemd/system/containerd.service.d/"
     state: directory
 
+- name: Create containerd max files
+  template:
+    dest: /etc/systemd/system/containerd.service.d/max-files.conf
+    src: etc/systemd/system/containerd.service.d/max-files.conf
+    mode: 0644
+
 - name: Create containerd proxy conf
   template:
     dest: /etc/systemd/system/containerd.service.d/http-proxy.conf
--- a/.github/workflows/vsphere-e2e.yaml
+++ b/.github/workflows/vsphere-e2e.yaml
@@ -84,4 +84,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.MESOSPHERECI_USER_TOKEN }}
           VSPHERE_SERVER: ${{ secrets.VSPHERE_SERVER }}
           VSPHERE_DATASTORE: ${{ secrets.VSPHERE_DATASTORE }}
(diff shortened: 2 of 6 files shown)


dkoshkin (approved):

mkarroqe (approved):

supershal (approved):

dlipovetsky:
When we upgraded containerd 1.4 to 1.6, the systemd unit changed the file descriptor limit from `1024768` to `infinity`. This PR effectively returns the limit to what it was in DKP 2.4 and earlier.

dlipovetsky (approved):

faiq:
Looking upstream this seems to be caused by [this PR](https://github.com/containerd/containerd/pull/4475/files) which sets the containerd limits to unlimited, users who were previously on 1.4.13 version of containerd provided by dkp 2.4 and previous had this limit set. Removing this limit caused our rook deployment to break and we also saw problem with HA proxy as well which had this limit as the root cause. We decided to bring this limit back for a couple of reasons

1. We don’t know if this can break existing users applications 
2. It breaks our application

dkoshkin:
@faiq I think we are good to merge this, the 2 other fixes look good:

* [RHSM vars](https://github.com/mesosphere/konvoy-image-builder/actions/runs/4325203880/jobs/7551054047#step:9:2455)
* [vSphere dry-run](https://github.com/mesosphere/konvoy-image-builder/actions/runs/4325203867/jobs/7551050473#step:9:3692)

ansible/roles/containerd/tasks/main.yaml
```diff
@@ -4,6 +4,12 @@
     path: "/etc/systemd/system/containerd.service.d/"
     state: directory
 
+- name: Create containerd max files
+  template:
+    dest: /etc/systemd/system/containerd.service.d/max-files.conf
+    src: etc/systemd/system/containerd.service.d/max-files.conf
+    mode: 0644
```
fatz, line 11: @faiq I started to rebase my molecule test PR #553 

it turns out that this part fails with Ubuntu. This is when `http_proxy` or `https_proxy`is not set and the first task is not run due to that. 

My suggestion:

Move this part after install includes and remove the conditions from directory creation

fatz: @faiq cherry-picking 137be77639a90cb75cdc3d55831404e02b85823a should help

PR #715, opened 2023-03-03
build: cache ansible collections in devkit image

supershal:
Installing ansible collection is the most time consuming step when building KIB container image. Caching it in devkit container should save around `1m30s` when building the image.

Tested locally.
```bash
❯ docker run --entrypoint '/bin/sh' -it --rm mesosphere/konvoy-image-builder:v2.2.1-15-ge02168df3248-amd64
~ # ls
ansible  images   packer
~ # ls /usr/share/ansible/collections/ansible_collections/
ansible                       ansible.posix-1.5.1.info      community
ansible.netcommon-5.0.0.info  ansible.utils-2.9.0.info      community.general-6.4.0.info
~ # ansible-galaxy collection list

# /usr/share/ansible/collections/ansible_collections
Collection        Version
----------------- -------
ansible.netcommon 5.0.0
ansible.posix     1.5.1
ansible.utils     2.9.0
community.general 6.4.0
```
E2E tests are passing too.

```release-note

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,15 +25,6 @@ RUN apk add --no-cache \
         py3-wheel \
     && pip3 install --no-cache-dir --requirement /tmp/requirements.txt \
     && rm -rf /root/.cache
-# Managing the below ansible dependencies is covered in docs/dev/ansible-modules.md
-RUN mkdir -p /usr/share/ansible/collections \
-    && ansible-galaxy \
-    collection install \
-    community.general \
-    ansible.netcommon \
-    ansible.posix \
-    ansible.utils \
-    -p /usr/share/ansible/collections
 
 ARG BUILDARCH
 # we copy this to remote hosts to execute GOSS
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -25,15 +25,6 @@ RUN apk add --no-cache \
         py3-wheel \
     && pip3 install --no-cache-dir --requirement /tmp/requirements.txt \
     && rm -rf /root/.cache
-# Managing the below ansible dependencies is covered in docs/dev/ansible-modules.md
-RUN mkdir -p /usr/share/ansible/collections \
```
dkoshkin, line 29: Any concerns with drifting versions for either Ansible or these collections?

supershal: version skew with ansible-core might be issue. We have this is a problem with current configuration as well. 

Documentation for each of these plugins only states that they are compatible with ansible core version X or higher. ex. https://docs.ansible.com/ansible/latest/collections/community/general/index.html
I think we can either
(1) add version constraint for ansible collections using https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#installing-an-older-version-of-a-collection
I check dependabot does not maintain galaxy collection bumps as of now. https://github.com/dependabot/dependabot-core/issues/3522 . we will have to maintain it manually.
We have to make sure that they are compatible with `ansible-core` version in `requirements.txt` file.
add a note in `requirements.txt` to ensure to check collection's versions when bumping `ansible-core` version.

(2) If we should always install latest version of the collections by installing collection every time we build KIB image instead of copying it from cached devkit image.
revert this change and remove installing collections in devkit image. 

wdyt?

dkoshkin: Sorry, I should have been clearer. I was more wondering about new version skew between kib image and devkit image?
It sounds like this PR is adding to that.

dkoshkin (approved):

faiq (approved):
I dont see these ansible collections changing that dramatically from the way that we're using them (very generically) 

also this saves an API call to github which we have constant rate limit battles with

PR #723, opened 2023-03-09
fix: allow building aws images using source_ami 

supershal:
Packer fails building AMI when user provides `source_ami` instead of `source_ami_filters`
Packer fails when evaluating empty ami_filter string: `""`
error:
```
2023/03/09 03:54:12 starting packer build
Error: Datasource.Execute failed: Couldn't parse ami filters: EOF
```
related upstream issue: https://github.com/hashicorp/packer/issues/11037
Fix:
- set the ami_filter to empty block `{}` when source_ami is provided.
- the packer ami datasource requires `owner` field. so setting it to `self` in case user does not provide it or explicitly set it to empty (`""`). see  https://developer.hashicorp.com/packer/plugins/builders/amazon/ebs#owners

This is a special usecase that our e2e tests does not cover.
I verified the fix by testing manually. 

*https://d2iq.atlassian.net/browse/D2IQ-96305

Packer uses HCL2 for its configuration language. This is same language used in Terraform. however, it is more restrictive when used with packer. This makes it difficult to write conditional blocks.
Other options considered:
- **use `count` to control block execution**. This is most preferred way in terraform HCL. This does not work with Packer.
error: `An argument named "count" is not expected here` 
reference: https://github.com/hashicorp/packer/issues/11483#issuecomment-1084970597

- **use dynamic blocks**
documentation: https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks
This solution works at some extent to render empty data resource. 
However the `source_ami_filter` block requires `owner` field which makes it difficult to completely omit the `data` resource for `ami`.

- **use local variables**
Packer does not support using local variables in `data` resource block. makes it difficult to create conditional variables.
Packer issue: https://github.com/hashicorp/packer/issues/11011
 
- **Use Go template** 
https://github.com/mesosphere/konvoy-image-builder/commit/9b53818503a7dcc4bdefdade42d34f2784a52548
We have used go template before. While this is a great solution, we will end up with mix of go templates and raw packer files. Raw packer files has advantage of testing them directly without getting rendered by KIB. which makes it easy to iterate when triaging bugs.


pkg/packer/manifests/aws/packer.pkr.hcl
```diff
@@ -316,14 +316,15 @@ variable "run_tags" {
 # https://www.packer.io/docs/datasources/amazon/ami
 data "amazon-ami" "autogenerated_1" {
   access_key = var.aws_access_key
-  filters = {
+  filters = var.source_ami != "" ? {} : {
```
dlipovetsky, line 319: HCL has a https://en.wikipedia.org/wiki/Ternary_conditional_operator? :scream:

msdolbey (approved):
👍 
Thanks! Let me know when merged, and I'll be able to check it out!
NO FILTER

dkoshkin (approved):
Tested locally passing source AMI
```
➜  konvoy-image-builder git:(shalin/fix-ami-source) ✗ ./bin/konvoy-image-wrapper build aws --overrides overrides/fips.yaml images/ami/rhel-84-fips.yaml --source-ami ami-07da62bb37a54c3b7
2023/03/09 21:03:57 writing new packer configuration to work/rhel-8.4-fips-1678395837-fZfQj
2023/03/09 21:03:57 starting packer build
amazon-ebs.kib_image: output will be in this color.

==> amazon-ebs.kib_image: Prevalidating any provided VPC information
==> amazon-ebs.kib_image: Prevalidating AMI Name: konvoy-ami-rhel-8.4-fips-1.25.4-fips.0-20230309210405
    amazon-ebs.kib_image: Found Image ID: ami-07da62bb37a54c3b7
    ...
```

pkg/packer/manifests/aws/packer.pkr.hcl
```diff
@@ -316,14 +316,15 @@ variable "run_tags" {
...
     architecture        = "x86_64"
     name                = var.ami_filter_name
     root-device-type    = "ebs"
     virtualization-type = "hvm"
   }
   most_recent = true
-  owners      = split(",", var.ami_filter_owners)
+  # owner is required field. default to "self" which will point to user's AWS account
```
dlipovetsky, line 326: nit
```suggestion
  # owners is required field. Default to "self" which will point to user's AWS account
```

dlipovetsky (approved):

PR #738, opened 2023-03-16
feat: do not copy stdin and set tty to false 

jonathanbeber:
This commit removes the instruction to copy the stdin in the docker wrapper and set tty to true is some cases. We don't need that since we don't expect KIB to read from stdin.

* https://jira.d2iq.com/browse/D2IQ-NOTCREATED

```release-note

```

### DIFF
--- a/cmd/konvoy-image-wrapper/cmd/wrapper.go
+++ b/cmd/konvoy-image-wrapper/cmd/wrapper.go
@@ -289,7 +289,6 @@ func (r *Runner) setupSSHAgent() {
 }
 
 func (r *Runner) dockerRun(args []string) error {
-	//nolint:gosec // running docker is inherently insecure
 	cmd := exec.Command(
 		"docker", "run",
 		"--interactive",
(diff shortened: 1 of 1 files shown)


cmd/konvoy-image-wrapper/cmd/wrapper.go
```diff
@@ -292,11 +289,10 @@ func (r *Runner) setupSSHAgent() {
 }
 
 func (r *Runner) dockerRun(args []string) error {
-	//nolint:gosec // running docker is inherently insecure
 	cmd := exec.Command(
 		"docker", "run",
 		"--interactive",
-		"--tty="+strconv.FormatBool(r.tty),
+		"--tty=false",
```
faiq, line 295: should we remove this flag in its entirety?

faiq (approved):
Looks good! Thank you and we got rid of one more dependency

fatz (approved):

fatz:
@faiq shall we plan a backport or better leave it to upcoming versions?

faiq:
I don’t think it needs a backport.

PR #739, opened 2023-03-16
fix: uses 80gb for all azure builds and removes extra

faiq:
This deletes the extra 4 and 5 partitions on azure. We need to explore alternatives to this hard coded approach 

```release-note

```

### DIFF
--- a/ansible/roles/providers/tasks/azure-redhat.yml
+++ b/ansible/roles/providers/tasks/azure-redhat.yml
@@ -25,21 +25,27 @@
       shell: dracut -f
 
     # https://d2iq.atlassian.net/browse/D2IQ-96417 removes extra 4 and 5 partitions so that growpart cloud-init module can expand the root filesystem to the rest of the disk.
-   - name: Remove partition number 4
-      parted:
-        device: /dev/sda
-        number: 1
-        state: absent
+    #
+    # [packer@pkrvmia9ljyrxtg ~]$ lsblk -oMOUNTPOINT,PKNAME -P | grep 'MOUNTPOINT="/"' | cut -d= -f3  | xargs
+    # sda
+    - name: Get Device that holds root filesystem
+      shell: lsblk -oMOUNTPOINT,PKNAME -P | grep 'MOUNTPOINT="/"' | cut -d= -f3  | xargs
+      register: rootdevice
 
-   - name: Remove partition number 5
+    - name: Loop over ansible block devices
       parted:
-        device: /dev/sda
-        number: 5
+        number: "{{ item }}"
         state: absent
+        device: "/dev/{{ rootdevice.stdout }}"
+      with_items:
+        - 4
+        - 5
 
     - name: unconditionally reboot the machine with all defaults
       ansible.builtin.reboot:
-  when: ansible_distribution_major_version|int >= 9
+  when:
+    - packer["image_publisher"] == "erockyenterprisesoftwarefoundationinc1653071250513" and packer["image_version"] == "9.1.20230215"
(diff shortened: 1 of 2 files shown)


ansible/roles/providers/tasks/azure-redhat.yml
```diff
@@ -24,6 +24,19 @@
     - name: refresh dracut
       shell: dracut -f
 
+    # https://d2iq.atlassian.net/browse/D2IQ-96417 removes extra 4 and 5 partitions so that growpart cloud-init module can expand the root filesystem to the rest of the disk.
+   - name: Remove partition number 4
```
faiq, line 28: consider using 

```
df / --output=source
Filesystem
/dev/sda3
```
and ansible_devices

faiq: consider scoping this down to only that azure image.

ansible/roles/providers/tasks/azure-redhat.yml
```diff
@@ -24,6 +24,19 @@
...
+      parted:
+        device: /dev/sda
+        number: 1
+        state: absent
```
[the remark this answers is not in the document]
faiq: much nicer. let's see if we can make this logic a bit smarter otherwise we can go with this.

pkg/packer/manifests/azure/packer.pkr.hcl
```diff
@@ -355,6 +355,7 @@ source "azure-arm" "kib_image" {
   managed_image_name                = local.managed_image_name
   managed_image_resource_group_name = var.resource_group_name
   os_type                           = "Linux"
+  os_disk_size_gb                   = 80
```
supershal, line 358: lets make this a packer variable with default to `80`

faiq: for sure, much cleaner. just wanted to capture it here before forgetting

ansible/roles/providers/tasks/azure-redhat.yml
dkoshkin: >consider scoping this down to only that azure image.

Yes please, lets only do this for Azure.

faiq: yeah, its already in the azure provider task, but we want to see if we could scope it down to that particular base image for rocky linux 9.1 @dlipovetsky is looking into how we can determine this using the metadata service and ansible facts

dlipovetsky: Ansible only has an ec2 module to query the metadata service. Although the metadata service is somewhat standardized across infrastructures, I'm not sure the module works on Azure.

We have the instance metadata on disk, in `/var/run/cloud-init/instance-data.json`.

I'm now looking for the appropriate field to read. I think it's `storageProfile.imageReference`.

ansible/roles/providers/tasks/azure-redhat.yml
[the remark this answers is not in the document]
dlipovetsky: Addressed

ansible/roles/providers/tasks/azure-redhat.yml
dlipovetsky: We were able to do this without much complexity by checking the metadata in Ansible.

dlipovetsky:
Workaround applied on Rocky: https://github.com/mesosphere/konvoy-image-builder/actions/runs/4450676362/jobs/7816421015?pr=739#step:8:2551

PR #749, opened 2023-03-20
fix: Retry containerd yum/rpm tasks up to 30s

dlipovetsky:
We found that some network failures lasted longer than 9s.

* https://d2iq.atlassian.net/browse/D2IQ-93681

I think exponential backoff would have been nice to use, but Ansible does not support it, despite a misleading example in the documentation https://github.com/ansible/ansible/issues/75794.

```release-note
```


faiq (approved):
More retries is definitely better. Sad ansible doesn’t support an exponential back off

PR #752, opened 2023-03-22
ci: update vsphere base template defaults

supershal:
update defaults to create kib bastion host to reflect current vSphere infrastrure


faiq (approved):
Thank you for updating this. I created a new machine using Jan's templates as well, so we can discuss what direction we want to go in later.

PR #760, opened 2023-03-24
FIX: cloud init on EL7

fatz:
When EL7 images come with cloud-init or have python3 installed KIB either overrides cloud-init files or uses the wrong python and sometimes even both.

This is due to missing guards in ansible.

The patch script also does not check which python cloud-init uses but will use python3 if its installed which is wrong in the case of cloud-init 19 on EL7

* https://jira.d2iq.com/browse/D2IQ-96572

```release-note

```

### DIFF
--- a/ansible/roles/providers/tasks/vmware.yml
+++ b/ansible/roles/providers/tasks/vmware.yml
@@ -12,122 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-- name: Install cloud-init packages
-  apt:
-    name: "{{ packages }}"
-    state: present
-    force_apt_get: true
-  vars:
-    packages:
-      - cloud-init
-      - cloud-guest-utils
-      - cloud-initramfs-copymods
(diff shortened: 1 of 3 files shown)


ansible/roles/providers/tasks/vmware.yml
```diff
@@ -12,122 +12,19 @@
...
+  changed_when: false
+
+- name: Set cloud-init version fact
+  set_fact:
+    cloud_init_version: "cloud_init_version_cmd.stdout"
+
+# when cloud-init 21.3 or higher is installed to not touch it
+- name: Cloud-init patches for cloud-init versions <21.3
+  include_tasks: vmware-cloudinit.yml
+  when: ( cloud_init_version == "" ) or (str(cloud_init_version) is version('21.3', '<'))
```
dkoshkin, line 27:
```suggestion
  when: ( cloud_init_version == "" ) or ((cloud_init_version | string) is version('21.3', '<'))
```

PR #714, opened 2023-03-03
ci: execute all e2e tests on PR label with one GHA workflow

supershal:
**What problem does this PR solve?**:
- Simplifies conditions to execute e2e jobs
- One job to control execution of all the e2e tests on a PR label
- Ensure devkit image is available on dockerhub before running tests
- Make it easy to navigate all e2e tests for current PR in one place in GHA UI

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-NUMBER

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #766, opened 2023-03-29
fix: pin ansible collection versions

supershal:
The latest ansible plugin `community.general` `6.5.0` breaks existing behavior of the `redhat_subscription` module. Issue: https://github.com/ansible-collections/community.general/issues/6258
As a result kib jobs fails when running RHEL builds in non-airgapped mode.
As a workaround, this PR pins ansible plugin versions.


supershal:
The fix for `main` branch is part of https://github.com/mesosphere/konvoy-image-builder/pull/714

dkoshkin (approved):
Thanks!

PR #716, opened 2023-03-03
feat: vSphere cloud-init / build command

fatz:
**What problem does this PR solve?**:
https://d2iq.atlassian.net/browse/D2IQ-93355

Support different setups by issuing cloud-init:
- support only public key given and expect ssh-agent heaving private key
- support neither public or private key given. Generate a key that will be used for authentication
- any username can be specified

Add the missing vsphere build command to have the same simple interface as with other clouds

```
bin/konvoy-image build vsphere --help
build and provision vsphere images

Usage:
  konvoy-image build vsphere <image.yaml> [flags]

Examples:
vsphere --datacenter dc1 --cluster zone1 --datastore nfs-store1 --network public --template=d2iq-base-templates/d2iq-base-CentOS-7.9 images/ami/centos-79.yaml

Flags:
      --cluster string               vSphere cluster to be used. Alternatively set host (REQUIRED)
      --containerd-version string    the version of containerd to install
      --datacenter string            vSphere datacenter (REQUIRED)
      --datastore string             datastore used to build and store the KIB image (REQUIRED)
      --dry-run                      do not create artifacts, or delete them after creating. Recommended for tests.
      --extra-vars strings           flag passed Ansible's extra-vars
      --folder string                deploy KIB image into this folder
  -h, --help                         help for vsphere
      --host string                  vSphere host to be used. Alternatively set cluster (REQUIRED)
      --kubernetes-version string    The version of kubernetes to install. Example: 1.21.6
      --network string               vSphere network to start KIB build; KIB expects DHCP on this network; Ensure the machine running KIB has access to this network (REQUIRED)
      --overrides strings            a comma separated list of override YAML files
      --packer-manifest string       provide the path to a custom packer manifest
      --packer-on-error string       [advanced] set error strategy for packer. strategies [cleanup, abort, run-cleanup-provisioner]
      --packer-path string           the location of the packer binary (default "packer")
      --resource-pool string         vSphere resource pool to be used to build KIB image
      --ssh-privatekey-file string   path to ssh private key which will be used to loginto the template
      --ssh-publickey string         SSH public key which will be deployed using cloud-init. Ensure to set ssh-privatekey-file or load the private key into ssh-agent
      --ssh-username string          Username to be used with the vSphere template
      --template string              Base template to be used. Can include folder. <templatename> or <folder>/<templatename> (REQUIRED)
      --work-dir string              path to custom work directory generated by the generate command

Global Flags:
      --color     enable color output (default true)
[...]

PR #800, opened 2023-05-09
feat: vsphere cloud init and command 

fatz:
What problem does this PR solve?:
This PR is cloning https://github.com/mesosphere/konvoy-image-builder/pull/716 due to a conflict resolving issue. This PR cherry-picks #716 commits on top of main


cmd/konvoy-image/cmd/vsphere.go
```diff
@@ -0,0 +1,139 @@
...
+		&vsphereArgs.Template,
+		"template",
+		"",
+		"Base template to be used. Can include folder. <templatename> or <folder>/<templatename> (REQUIRED)",
+	)
+	fs.StringVar(
+		&vsphereArgs.Datacenter,
+		"datacenter",
+		"",
+		"vSphere datacenter (REQUIRED)",
```
faiq, line 77: These aren't required 

```
bin/konvoy-image-wrapper with args [build vsphere images/ova/rhel-79.yaml --dry-run --overrides=overrides/offline.yaml --overrides=packer-ova-offline-override.yaml]
```

They can be in the overrides or passed in through these flags.

fatz: why is datacenter not required? 

I don't think it will run without and if we have some defaults somewhere we rather should delete these defaults as customer resources will definitly be named differently

fatz: https://github.com/mesosphere/konvoy-image-builder/blob/main/pkg/packer/manifests/vsphere/packer.pkr.hcl#L179-L182C2

we could name the additional variable here. But no matter how its set datacenter is needed

faiq: > why is datacenter not required?

> I don't think it will run without and if we have some defaults somewhere we rather should delete these defaults as customer resources will definitly be named differently

we do not have defaults anywhere. this is being passed in through an override like in the command I posted 

```
--overrides=packer-ova-offline-override.yaml
```

It wouldn't be good to force the user to only use this flag when it was being passed in previously via override.

Therefore, setting the flag is not required. The value of datacenter is required. I hope this clears things up

fatz: i see... 

Its probably just my personal perspective on these things...

So the `--overrides` as well as environment variable are just another way to full fill the requirement here. Which lets me still think that data is required. 

- So we agree that data center is a required input.
- KIB offers multiple ways to inject this data

Having the sub-command with this PR makes the step of having the override file completely obsolete. 

The huge benefit is its self-documenting.

When we drop the required mark from `--datacenter` we must add additional documentation that now either `--overrides` or env `VSPHERE_DATACENTER` are being required. Which IMO makes it more complicated to understand.

> It wouldn't be good to force the user to only use this flag when it was being passed in previously via override
This sub command was not existent so we do not force anything as plain `build` stays absolutely the same

faiq: > When we drop the required mark from --datacenter we must add additional documentation that now either --overrides or env VSPHERE_DATACENTER are being required. Which IMO makes it more complicated to understand.

I think we can add this in the flag description "must be passed in via this flag, override file, or included in ova template yaml file"

fatz: So I just looked into the other sub-commands. They just completely miss what is required and what is optional.

While neither environment variables as well as the override stuff is documented in the command itself. You have to dig deep into our documentation to understand this features.

While many things in public clouds could be assumed this won't be possible in customer driven environments like vsphere where every single dependent resource is custom defined.


I just tried to add some additional notes into the flags but it just makes things complicated.



So can you tell me why documenting that this flag is required is a problem at all?

Leaving it out makes the documentation worse which is something we should not even consider due to this process already being over-complicated.


Frankly I would even question if the `<image.yaml>` input could be left out. In vsphere it only has use-less information which gets forwarded into ansible while ansible by its nature knows already about it. 
The only information left in that file is IMO not really important as it just controls some metadata for the template

fatz: Great idea thank you !

cmd/konvoy-image/cmd/vsphere.go
```diff
@@ -0,0 +1,146 @@
...
+		"",
+		"Base template to be used. Can include folder. <templatename> or <folder>/<templatename>."+
+			"Required value: you can pass the template name through an override file, or image definition file.",
+	)
+	fs.StringVar(
+		&vsphereArgs.Datacenter,
+		"datacenter",
+		"",
+		"The vSphere datacenter."+
+			"Required value: you can pass the datacenter name through an override file, or image definition file.",
```
charIesdu, line 79: Change to `"Required value: you can pass the datacenter name through an override file or image definition file."` here and every other line that has the verbiage. 

Add a space character anywhere it says `Required value`

--host and --template also need spaces per @JohnB64-JBV. Review the rest of the code to double-check.

faiq: Fixed!

faiq (approved):
5/5 changes!

fatz:
@faiq WDYT should we also rename the `images/ova` folder? 

- for aws its also named after the provider and not the image format (ami)
- we could  `git mv images/ova images/vsphere` and add a softlink for ova for backwards compatibility

this will make the naming more unified and so subcommand and folder are matching.

I can add that patch but did not want to change the PR after you approved

supershal:
> should we also rename the images/ova folder
@fatz as part of refactoring KIB, we are considering different approach to handle this. I think we can skip such feature enhancements as they might not be transferred as it is in new design.

cmd/konvoy-image/cmd/vsphere.go
supershal: > I would even question if the <image.yaml> input could be left out
We are moving out <image.yaml> in new design. I DM you details.

PR #805, opened 2023-05-15
ci: changed runner labels 

shaeed-noordeen:
There were some workflows that still had the wrong labelled runners or were using githubs runners, we need to change them to org-wire-runners so we have a more coherent understanding of our infrastructure. This pr will be used to understand which workflows require which size or if they need kind cluster. Please check the the runners in the change made and use [this](https://github.com/mesosphere/github-action-runners-deployment#how-to-add-an-org-wide-runner-to-your-repos-workflow) as a criteria. Once the sizes have been confirmed the pr we can merge.

Which issue(s) does this PR fix?:

https://jira.d2iq.com/browse/D2IQ-97191

```release-note

```

### DIFF
--- a/.github/workflows/azure-e2e.yaml
+++ b/.github/workflows/azure-e2e.yaml
@@ -19,7 +19,9 @@ jobs:
         - "rocky 9.1"
         buildConfig:
         - "basic"
-    runs-on: self-hosted
+    runs-on: 
+    - self-hosted
+    - small
     continue-on-error: false
     steps:
       - name: Checkout konvoy-image-builder repository
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ permissions:
 jobs:
   super-linter:
     runs-on: 
-    - self-hosted§
+    - self-hosted
     - small
     steps:
       - name: Checkout konvoy-image-builder repository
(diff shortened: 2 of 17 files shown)


.github/workflows/azure-e2e.yaml
```diff
@@ -19,7 +19,9 @@ jobs:
         - "rocky 9.1"
         buildConfig:
         - "basic"
-    runs-on: self-hosted
+    runs-on: 
+    - self-hosted
+    - small
```
dkoshkin, line 24: What size is selected if not set?

.github/workflows/lint.yml
```diff
@@ -15,7 +15,9 @@ permissions:
 
 jobs:
   super-linter:
-    runs-on: ubuntu-latest
+    runs-on: 
+    - self-hosted§
```
dkoshkin, line 19:
```suggestion
    - self-hosted
```

.github/workflows/vsphere-e2e.yaml
```diff
@@ -30,7 +30,9 @@ jobs:
           buildConfig: offline-fips
         - os: "rocky 9.1"
           buildConfig: offline-fips
-    runs-on: self-hosted
+    runs-on: 
+- self-hosted
+- small
```
dkoshkin, line 35:
```suggestion
    runs-on: 
    - self-hosted
    - small
```

supershal:
@shaeed-noordeen The jobs are not getting any self-hosted runners. can you please take a look.

PR #808, opened 2023-05-16
fix: adds missing package to address sles issue

faiq:
 Adds apparmor abstractions package which allows the upgrade control-plane command to return successfully

* https://d2iq.atlassian.net/browse/D2IQ-96645

```release-note

```

### DIFF
--- a/ansible/roles/containerd/tasks/suse.yaml
+++ b/ansible/roles/containerd/tasks/suse.yaml
@@ -4,10 +4,6 @@
   zypper:
     name: containerd
     state: absent
-  register: result
-  until: result is success
-  retries: 15
-  delay: 60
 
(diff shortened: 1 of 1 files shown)


ansible/roles/containerd/tasks/suse.yaml
```diff
@@ -37,11 +30,13 @@
...
-    name: libseccomp2
+    name:
+      - libseccomp2
+      - apparmor-abstractions
     state: present
   register: result
   until: result is success
-  retries: 15
-  delay: 60
+  retries: 3
```
supershal, line 41: we intentionally added higher retries to go around CI flakes with SLES

dkoshkin (approved):

PR #836, opened 2023-06-15
feat: Upgrade kubernetes version to v1.26.6 and CRI tools

bcmendoza:
Updates crictl to 1.26.0

* https://d2iq.atlassian.net/browse/D2IQ-96607

```release-note

```


dkoshkin:
I expect some of the AWS tests to fail here because the Kubernetes v1.26.3 bundles were built with the older version, you will need to make this change in the same PR as the Kubernetes bump.

dkoshkin (approved):

PR #861, opened 2023-08-03
fix: bumps ansible to remove unneeded ssh arguments

faiq:
bumps the ansible packer version so we can remove the SSH Extra Args. 

Also includes a fix for a conditional check that unsubscribes the rhel system. Previously, if the system was not registered this would fail.

```release-note

```


dkoshkin (approved):
LGTM! If it lets you, please rebase and merge since we may want to backport these commits separately

PR #866, opened 2023-08-07
fix: aws cli issue  flatcar

faiq:

```release-note

```

### DIFF
--- a/ansible/roles/providers/tasks/aws.yml
+++ b/ansible/roles/providers/tasks/aws.yml
@@ -19,11 +19,27 @@
   when:
     - ansible_os_family == "Debian"
 
+- name: install cython for flatcar
+  pip:
+    name: "cython<3.0.0"
+    executable: pip3
+    extra_args: "wheel"
+  when:
+    - ansible_distribution == "Flatcar"
+
+- name: install pyyaml
+  pip:
+    name: "pyyaml==5.4.1"
+    executable: pip3
+    extra_args: "--no-build-isolation"
+  when:
+    - ansible_distribution == "Flatcar"
+
 - name: install aws clients
   pip:
     name: "awscli"
     executable: pip3
-    extra_args: "{{ '--no-index --find-links=' + pip_packages_remote_filesystem_repo_path if offline_mode_enabled }}"
+    extra_args: "{{ '--no-cache-dir --no-index --find-links=' + pip_packages_remote_filesystem_repo_path if offline_mode_enabled }}"
   when:
     - ansible_distribution != "Amazon"
     - ansible_os_family != "Suse"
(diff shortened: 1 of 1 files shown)


ansible/roles/providers/tasks/aws.yml
```diff
@@ -19,11 +19,27 @@
...
+  pip:
+    name: "pyyaml==5.4.1"
+    executable: pip3
+    extra_args: "--no-build-isolation"
+  when:
+    - ansible_distribution == "Flatcar"
+
 - name: install aws clients
   pip:
     name: "awscli"
```
faiq, line 40: originally i tried pinning the version to 1.29.9 like we have in requirements-devkit.txt but this didn't work.

i then tried to see if i could use a newer pypy version https://www.pypy.org/download.html but this didn't work either.

ansible/roles/providers/tasks/aws.yml
```diff
@@ -19,11 +19,27 @@
...
+  when:
+    - ansible_distribution == "Flatcar"
+
+- name: install pyyaml
+  pip:
+    name: "pyyaml==5.4.1"
+    executable: pip3
+    extra_args: "--no-build-isolation"
+  when:
+    - ansible_distribution == "Flatcar"
```
supershal, line 36: we should loop over using `with_items` in a single task.

ansible/roles/providers/tasks/aws.yml
```diff
@@ -19,11 +19,27 @@
...
+    executable: pip3
+    extra_args: "wheel"
+  when:
+    - ansible_distribution == "Flatcar"
+
+- name: install pyyaml
+  pip:
+    name: "pyyaml==5.4.1"
+    executable: pip3
+    extra_args: "--no-build-isolation"
```
supershal, line 34: can you please brief documentation why extra_args `wheel` and `--no-build-isolation` for future reference.
the doc for `--no-build-isolation` is `Disable isolation when building a modern source`. but I am not sure what it means.

ansible/roles/providers/tasks/aws.yml
faiq: They have different extra args so we can’t

faiq: Wait might have figured something out

ansible/roles/providers/tasks/aws.yml
faiq: I'm not sure why, I copied this answer from stack overflow and tested it. 

I added a comment linking to the answer.

supershal: Thank you.

supershal (approved):
LGTM. lets wait for e2e tests to pass.

PR #871, opened 2023-08-18
[release-2.2] ci: use vsphere base templates from d2iq-base-templates vsphere folder

dkoshkin:
uses d2iq base templates to build KIB OVA.
When we disable the ssh_agent based authentication, the packer code uses temporary keypair generated during provisioning.
Existing logic that manages this is at: https://github.com/mesosphere/konvoy-image-builder/blob/main/pkg/packer/manifests/vsphere/packer.pkr.hcl#L378-L405

### DIFF
--- a/images/ova/rhel-79.yaml
+++ b/images/ova/rhel-79.yaml
@@ -13,7 +13,7 @@ packer:
   insecure_connection: "false"
   network: ""
   resource_pool: ""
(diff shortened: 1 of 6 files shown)


dkoshkin:
Fixed in https://github.com/mesosphere/konvoy-image-builder/pull/901, these changes were not enough anyways.

PR #872, opened 2023-08-18
[release-2.5] ci: use vsphere base templates from d2iq-base-templates vsphere folder

dkoshkin:
uses d2iq base templates to build KIB OVA.
When we disable the ssh_agent based authentication, the packer code uses temporary keypair generated during provisioning.
Existing logic that manages this is at: https://github.com/mesosphere/konvoy-image-builder/blob/main/pkg/packer/manifests/vsphere/packer.pkr.hcl#L378-L405


supershal (approved):
Thank you back porting and testing the fix.

PR #886, opened 2023-08-29
fix: start usr-libexec.mount to also support pre-provisioned

dkoshkin:
A pre-provisioned machine never gets restarted after running KIB, so this systemd unit never gets started.
Saw that `kube-controller-manager` in Konvoy e2e test was failing to start because `/user/libexec/kubernetes` was still read only.
Tested locally and with this change the pre-provisioned cluster came up.

* https://d2iq.atlassian.net/browse/D2IQ-98654

```release-note

```


supershal (approved):
Thank you for testing this end to end.

PR #888, opened 2023-08-30
feat: Upgrade Kubernetes version to 1.27.5

bcmendoza:
K8s upgrade

* https://d2iq.atlassian.net/browse/D2IQ-97979

According to https://github.com/kubernetes/kubernetes/blob/v1.27.5/cluster/gce/gci/configure.sh#L35, K8s 1.27.5 is still using crictl 1.26.0, so no need to upgrade.

```release-note

```

### DIFF
--- a/.github/workflows/aws-e2e.yaml
+++ b/.github/workflows/aws-e2e.yaml
@@ -12,7 +12,7 @@ jobs:
   rune2e:
     strategy:
       fail-fast: false
-      max-parallel: 10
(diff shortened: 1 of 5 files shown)


dkoshkin:
@bcmendoza please update the kubeadm config to use the newer API, see https://github.com/mesosphere/dkp-release/commit/1224ff0d62dbfdde192c023b7875abe151233d79

PR #912, opened 2023-09-14
chore: download pip packages from downloads.d2iq.com 

supershal:
- Directly downloads pip packages from downloads.d2iq.com
- refactor the magefile code to download airgap artifacts


supershal:
All tets except Nvidia passed. Nvidia tests are failing because of kernel header mismatch. looking into it.

PR #930, opened 2023-10-13
fix: use bin/konvoy-image-wrapper now to fix podman tests 

faiq:
**What problem does this PR solve?**:

**Which issue(s) does this PR fix?**:

* https://jira.d2iq.com/browse/D2IQ-NUMBER

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #931, opened 2023-10-17
fix: skip building podman binary on release

dkoshkin:
The release job [was failing here](https://github.com/mesosphere/konvoy-image-builder/actions/runs/6552028990/job/17796325263). I suspect its due the recent changes in [.goreleaser.yaml](https://github.com/mesosphere/konvoy-image-builder/pull/930/files#diff-42e26dc67aed8aa3edb2472b4403288c1699fb6dc47419b9a475f0f224fe4689R38) and its overriding the same `konvoy-image` binary. Since this new build spec is only needed in a specific e2e test I moved it out to a different (non-default) config file and reverted .goreleaser.yaml to what it was before.

```release-note

```

### DIFF
--- a/.github/workflows/podman-aws-e2e.yaml
+++ b/.github/workflows/podman-aws-e2e.yaml
@@ -53,7 +53,7 @@ jobs:
       - name: Build snapshot
         run:  |-
           make build-for-podman
-          goreleaser build --id konvoy-image-wrapper-for-podman --clean --skip-validate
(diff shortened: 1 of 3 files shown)


dkoshkin:
Running e2e test in https://github.com/mesosphere/konvoy-image-builder/actions/runs/6553913028/job/17800032651

supershal (approved):
Thanks for the fix that can work for both usecases

PR #1064, opened 2024-04-19
D2IQ-100548: add missing .aws/config file mount - main

fatz:
`.aws/config` is being ignored by KIB. This is an issue especially when custom config like `aws configure sso` is being used or special profiles.

Due to the D2iQ tooling relied completely on .aws/credentials this was never noticed before.

* https://jira.nutanix.com/browse/D2IQ-100548

`.aws/config` is an important part of SDK config. We completely forgot about it

No change expected. If the file is not provided or existing its should be ignored


supershal (approved):
Thanks for fixing this

PR #1072, opened 2024-05-06
feat: removes rhel 7.9 rhel 8.4 and centos 7.9

faiq:

Centos 7.9 is EOL in June this PR drops all support for it.

```release-note

```


jimmidyson (commented):
According to https://access.redhat.com/support/policy/updates/errata/#Life_Cycle_Dates technically I guess customers could still be using RHEL7 if they pay for extended support until 2028. I agree we should get rid of this ASAP but can we just confirm with product that we have no customers using this?

kevinantonythomas (approved):

supershal (approved):

PR #1081, opened 2024-05-17
fix: clean yum cache after adding repos 

supershal:
- Always clean yum cache after adding kubernetes repositories

Once we add a repo to the yum cache, we need to update the yum caches so that the packages from the repo can be downloaded.
when we install ansible packages we do use `update_cache: yes` option. https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html#parameter-update_cache
however it will clean the cache only if it is expired. So the new packages are not getting downloaded automatically even after ansible invokes update cache.
This PR forces yum to clean all yum caches including repo mirror list so that package list can be refreshed automatically. 

* https://jira.nutanix.com/browse/NCN-100851

### DIFF
--- a/ansible/roles/repo/tasks/redhat.yaml
+++ b/ansible/roles/repo/tasks/redhat.yaml
@@ -119,3 +119,6 @@
   until: konvoy_repo_installation_rpm is success
   retries: 5
   delay: 6
+
+- name: clean yum caches
+  shell: yum clean all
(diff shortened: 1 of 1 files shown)


dkoshkin (approved):
Can you please add a description in the PR why this is needed.

faiq (approved):

PR #1084, opened 2024-05-20
chore: bump dynamic-credential-provider to v0.5.3

supershal:
bumps dynamic credentials provider to v0.5.3

### DIFF
--- a/ansible/group_vars/all/defaults.yaml
+++ b/ansible/group_vars/all/defaults.yaml
@@ -59,7 +59,7 @@ aws_images: []
 
 extra_images: []
 
-containerd_supplementary_images: ["ghcr.io/mesosphere/toml-merge:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.5.0"]
+containerd_supplementary_images: ["ghcr.io/mesosphere/toml-merge:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.5.3"]
 
 containerd_base_url: https://packages.d2iq.com/dkp/containerd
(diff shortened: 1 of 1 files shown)


ansible/group_vars/all/defaults.yaml
```diff
@@ -59,7 +59,7 @@ aws_images: []
 
 extra_images: []
 
-containerd_supplementary_images: ["ghcr.io/mesosphere/toml-merge:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.5.0"]
+containerd_supplementary_images: ["ghcr.io/mesosphere/toml-merge:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0", "ghcr.io/mesosphere/dynamic-credential-provider:v0.5.3"]
```
jimmidyson, line 62: Why do we need both `ghcr.io/mesosphere/dynamic-credential-provider:v0.2.0` and `v0.5.3`?

supershal: I honestly does not have context about this. @dkoshkin any idea?

dkoshkin: This was the PR that added `v0.5.0` https://github.com/mesosphere/konvoy-image-builder/pull/1019
Konvoy is still using `v0.2.0` and the old API https://github.com/mesosphere/konvoy2/blob/051fdf94628a09028d0c4befe8d27b04876d37f7/pkg/capi/generator/templates/dynamic-credential-provider-config.yaml.gotmpl#L12

I think we should keep `v0.2.0` like you did here and cut another ticket to do the API migration in Konvoy.

dkoshkin (approved):

ansible/group_vars/all/defaults.yaml
supershal: Thank you for the context

PR #1096, opened 2024-05-29
ci: Use activation key and org ID instead of username and password 

dlipovetsky:
We won't be able to use the username/password in CI in the future, so use the activation key and org ID instead.

```release-note

```


faiq (approved):

dlipovetsky:
I also updated the vSphere readme, and env vars passed to docker.

PR #1099, opened 2024-06-04
fix: Enable unprivileged ports sysctl in containerd config

dkoshkin:
Spotted this issue again when testing vSphere install/upgrade with Kubernetes v1.29.5.

Similar to https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/645, but solving it here to handle the more generic case of so it can be picked up during upgrades without needing to add complicated `preKubeadmCommands` update logic in Konvoy2.

This enabled pods to run as non-root and bind to privileged ports as long as they have the necessary capability, `CAP_NET_BIND_SERVICE` added.

This fixes an issue on AWS when bringing up coredns which binds to port 53 but runs as an unprivileged user.

Overall this is a net security improvement for clusters, meaning users can stop giving too many privileged to pods - see
kubernetes/kubernetes#102612 for discussion.

* https://jira.nutanix.com/browse/NCN-100169

Ran the konvoy2 e2e test https://github.com/mesosphere/konvoy2/actions/runs/9370800641

```release-note

```

### DIFF
--- a/ansible/roles/config/templates/config.toml.tmpl
+++ b/ansible/roles/config/templates/config.toml.tmpl
@@ -71,6 +71,8 @@ imports = ["/etc/containerd/conf.d/*.toml"]
     restrict_oom_score_adj = false
     max_concurrent_downloads = 3
     disable_proc_mount = false
+    enable_unprivileged_ports = true
+    enable_unprivileged_icmp = true
     [plugins."io.containerd.grpc.v1.cri".containerd]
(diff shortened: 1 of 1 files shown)


dkoshkin:
Ran the vSphere build https://github.com/mesosphere/konvoy-image-builder/actions/runs/9370058290

faiq (approved):

PR #1116, opened 2024-07-09
fix: Make image load compatible with podman 4.x

dlipovetsky:
Remove workaround for podman 3.x image load behavior, because the workaround is incompatible with podman 4.x.

Also:
- Use Ubuntu 24.04 runner to run podman e2e test. This runner includes podman 4.x, which we require. 
- Use Rocky 9.1 to run podman e2e test. Previously, we used CentOS 7.9, but we cannot use it, because we dropped support for it recently (#1072), as it went EOL.

* https://jira.nutanix.com/browse/NCN-101343

```release-note

```

### DIFF
--- a/.github/workflows/podman-aws-e2e.yaml
+++ b/.github/workflows/podman-aws-e2e.yaml
@@ -58,9 +58,9 @@ jobs:
           make build-for-podman
           goreleaser build --config=.goreleaser-podman-e2e.yml --clean --skip-validate
 
-      - name: Run E2E test for AWS centos 7.9 using podman
+      - name: Run E2E test for AWS Rocky 9.1 using podman
         run: |-
-          KIB_CONTAINER_ENGINE=podman dist/konvoy-image-wrapper-for-podman_linux_amd64_v1/konvoy-image build aws images/ami/centos-79.yaml --dry-run
+          KIB_CONTAINER_ENGINE=podman dist/konvoy-image-wrapper-for-podman_linux_amd64_v1/konvoy-image build aws images/ami/rocky-91.yaml --dry-run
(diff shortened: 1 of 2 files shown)


.github/workflows/podman-aws-e2e.yaml
```diff
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   rune2e:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
faiq, line 13:
```suggestion
    # we are using ubuntu 24.04 because it come with version 4.9.3 of podman https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
    runs-on: ubuntu-24.04
```

faiq (approved):

.github/workflows/podman-aws-e2e.yaml
dlipovetsky: Thanks, good idea! I wrote a slightly different comment in https://github.com/mesosphere/konvoy-image-builder/pull/1116/commits/9df4f6401dadfbb2d3b82db38cb293e0a2c06f40.

dkoshkin (approved):

dlipovetsky:
Thanks for your reviews. I'm going to restructure the changes slightly, to make it clear that we're making a change to the product, as well as CI. I won't change the code in any way. I'll force push.

dlipovetsky:
For reference, the podman e2e test passes: https://github.com/mesosphere/konvoy-image-builder/actions/runs/9877813263/job/27280651164?pr=1116

PR #1136, opened 2024-08-05
ci: fetch kernel headers only for GPU tests

faiq:
Adds a gate around fetching kernel headers with a note about where to modify the file in the help text of the flag. Also updates CI to use this new flag.

```release-note

```

### DIFF
--- a/cmd/konvoy-image-wrapper/cmd/create-package-bundle.go
+++ b/cmd/konvoy-image-wrapper/cmd/create-package-bundle.go
@@ -82,6 +82,7 @@ func (r *Runner) CreatePackageBundle(args []string) error {
 		eusReposFlag          bool
 		outputDirectoy        string
 		containerImage        string
+		fetchKernelHeaders    bool
 	)
 	flagSet := flag.NewFlagSet(createPackageBundleCmd, flag.ExitOnError)
 	flagSet.StringVar(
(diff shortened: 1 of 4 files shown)


cmd/konvoy-image-wrapper/cmd/create-package-bundle.go
```diff
@@ -108,9 +109,24 @@ func (r *Runner) CreatePackageBundle(args []string) error {
...
+	)
+	flagSet.StringVar(
+		&containerImage,
+		"container-image",
+		"",
+		"A container image to use for building the package bundles",
+	)
+	flagSet.BoolVar(
+		&fetchKernelHeaders,
+		"fetch-kernel-headers",
```
supershal, line 126: I am wondering if we can get away with some generic solution and documentation about guiding users when creating bundle for GPU with their matching kernel.
For ex.
We can have generic flag `--additional-packages <Comma_sepeareted_list_of_rpms>` which then will be parsed with go template in `packages.txt.gotmpl` .

Document that they must download `kernel-headers-<version>` and `kernel-devel-<version>` using `--additional-packages kernel-headers-<version>, kernel-devel-<version>` . and provide instructions about how to find <version> for their base OS?

We modify our tests to include versions that matches with AMI and document what we tested?
Wdyt?

faiq: IMO the biggest issue with the feature of modifying the packages.txt.gotmpl is the lack of documentation around being able to modify it. 

I think adding another flag adds more complexity for the end user and in the code base as well, when there is already a way to modify this list.

I think we should consider this as a feature if customers are not happy with a documentation approach.

faiq: JIRA for docs here https://jira.nutanix.com/browse/NCN-101854

supershal (approved):

dkoshkin (approved):

PR #1156, opened 2024-08-20
[wip]feat: changes to support network restricted environments when creating RHEL OS bundles

some-things:

I don't have permissions to push images to test a3512a5b1335bc4425363b6c3e0359a8ccd2258c. Any help there would be greatly appreciated.

```release-note

```


bundles/redhat8.6/bundle.sh.gotmpl
```diff
@@ -13,57 +13,94 @@ RHSM_ACTIVATION_KEY=${RHSM_ACTIVATION_KEY:-""}
...
+fi
+
+if [[ -f /etc/yum.repos.d/user-repos.repo ]]; then
+  USER_REPOS="$(awk -F '[][]' '/^\[.*\]/ {print $2}' /etc/yum.repos.d/user-repos.repo | paste -sd, -)"
+  if [[ -n "${ENABLED_REPOS}" ]] && [[ -n "${USER_REPOS}" ]]; then
+      ENABLED_REPOS="${ENABLED_REPOS},${USER_REPOS}"
+  elif [[ -z "${ENABLED_REPOS}" ]] && [[ -n "${USER_REPOS}" ]]; then
+      ENABLED_REPOS="${USER_REPOS}"
+  fi
+fi
```
some-things, line 83: This requires users to define their rpm repos in `repo-templates/user-repos.repo`. We will need to document this workflow.

While KIB CLI automatically adds any repo files in `repo-templates/` to `/etc/yum.repos.d`, we need to selectively enable/disable repos that already exist there (e.g., `/etc/yum.repos.d/ubi.repo`). We could only parse files other than this one, but maybe it is simpler to just have one file that users modify and we look for.

Open to any feedback.

bundles/redhat8.6/bundle.sh.gotmpl
```diff
@@ -13,57 +13,99 @@ RHSM_ACTIVATION_KEY=${RHSM_ACTIVATION_KEY:-""}
...
```
faiq, line 88: I can think of another way of handling this. not a huge fan of usig awk this way

some-things: @faiq I changed this in 6809b18. Let me know what you think.

bundles/redhat8.8/bundle.sh.gotmpl
```diff
@@ -13,57 +13,99 @@ RHSM_ACTIVATION_KEY=${RHSM_ACTIVATION_KEY:-""}
...
+  --setopt=skip_missing_names_on_install=False -x \*i686 --archlist=x86_64,noarch $(<reqs.txt)
 #shellcheck disable=SC2046
-yumdownloader --setopt=skip_missing_names_on_install=False -x \*i686 --archlist=x86_64,noarch --resolve --disablerepo=*  --enablerepo=kubernetes,rhel-8-for-x86_64-baseos-eus-rpms,codeready-builder-for-rhel-8-x86_64-rpms,rhel-8-for-x86_64-appstream-rpms $(< packages.txt)
+yumdownloader --disablerepo=* --enablerepo="${ENABLED_REPOS}" \
+  --setopt=skip_missing_names_on_install=False -x \*i686 --archlist=x86_64,noarch \
+  --resolve $(<packages.txt)
 rm packages.txt reqs.txt
-curl https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm -o amazon-ssm-agent.rpm
 createrepo -v .
-chown -R 1000:1000 repodata/
```
faiq, line 66: why is this removed?

faiq: removed from main.

faiq:
I'm going to break this pr into smaller ones for release-2.9

regarding the satellite and skip subscription manager features open lets consider doing this in main first.

PR #1190, opened 2024-09-27
feat: add Oracle Linux 9.4 vSphere build configuration

dkoshkin:
This adds new configuration to build Oracle Linux 9.4 vSphere templates.

* https://jira.nutanix.com/browse/NCN-102630

✅ I built a template [here](https://github.com/mesosphere/konvoy-image-builder/actions/runs/11076002535) and then ran an install test [here](https://github.com/mesosphere/konvoy2/actions/runs/11076450767/job/30779936267#step:5:6702).
Will wait for a PR to support building the air-gapped bundle and then we can run the air-gapped tests there.

```release-note

```

### DIFF
--- a/.github/workflows/release-vsphere-template.yaml
+++ b/.github/workflows/release-vsphere-template.yaml
@@ -27,6 +27,10 @@ jobs:
             buildConfig: "basic"
           - os: "rocky 9.1"
             buildConfig: "offline"
+          - os: "oracle 9.4"
+            buildConfig: "basic"
+          - os: "oracle 9.4"
+            buildConfig: "fips"
           - os: "flatcar"
             buildConfig: "basic"
     runs-on:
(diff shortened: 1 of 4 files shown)


manoj-nutanix (approved):

.github/workflows/release-vsphere-template.yaml
```diff
@@ -27,6 +27,10 @@ jobs:
             buildConfig: "basic"
           - os: "rocky 9.1"
             buildConfig: "offline"
+          - os: "oracle 9.4"
+            buildConfig: "basic"
+          - os: "oracle 9.4"
+            buildConfig: "fips"
```
manoj-nutanix, line 33: which manifests is it using? konvoy2 repo doesn't contain FIPS manifests for oracle 9.4 atm

dkoshkin: The FIPS overrides just tell the CLI to install the FIPS packages, then Konvoy uses the manifests to compare the installed packages vs to what it expects.

manoj-nutanix: thanks for clarifying.

PR #1202, opened 2024-10-09
fix: remove appstream code as its not needed

faiq:

```release-note

```

### DIFF
--- a/ansible/roles/gpu/tasks/nvidia-gpu-RHEL.yaml
+++ b/ansible/roles/gpu/tasks/nvidia-gpu-RHEL.yaml
@@ -1,11 +1,13 @@
 ---
 - name: enable RHEL 8 repos
-  shell: subscription-manager repos --enable=rhel-8-for-x86_64-baseos-eus-rpms 
+  shell: subscription-manager repos --enable=rhel-8-for-x86_64-baseos-eus-rpms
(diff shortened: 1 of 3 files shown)


ansible/roles/gpu/tasks/nvidia-gpu-RHEL.yaml
```diff
@@ -1,11 +1,13 @@
 ---
 - name: enable RHEL 8 repos
-  shell: subscription-manager repos --enable=rhel-8-for-x86_64-baseos-eus-rpms 
+  shell: subscription-manager repos --enable=rhel-8-for-x86_64-baseos-eus-rpms
   when:
     - rhelorg != None or rheluser != None
     - ansible_distribution_major_version == '8'
+    - ansible_distribution_version is version('8.10', '<')
```
supershal, line 7: why we need this check for `8.10` ?

supershal (approved):

ansible/roles/gpu/tasks/nvidia-gpu-RHEL.yaml
faiq: EUS repos are only needed for EUS operating systems - 8.10 is a long term supported OS so we don't need it.

PR #1217, opened 2024-10-30
fix: use Rocky 9 without a minor version for Azure

dkoshkin:
Rocky deprecates versions as soon as the next minor is released https://wiki.rockylinux.org/rocky/version/#current-supported-releases.
Dropping the minor version in Azure to avoid referencing a removed image in the future. 

```release-note

```

### DIFF
--- a/images/azure/rocky-9.yaml
+++ b/images/azure/rocky-9.yaml
@@ -1,5 +1,5 @@
 ---
-build_name: "rocky-9-az"
+build_name: "rocky-9"
 packer_builder_type: "azure"
 python_path: ""
 packer:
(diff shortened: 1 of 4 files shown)


images/azure/rocky-9.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-build_name: "rocky-93-az"
+build_name: "rocky-9-az"
 packer_builder_type: "azure"
 python_path: ""
 packer:
```
faiq, line 5: should image_version be latest? or we want to leave it as is

dkoshkin: 🤦 yep, fixed.

supershal (approved):

PR #1235, opened 2024-12-02
fix: allows ipv6 to be disabled in goss checks

faiq:

Some customers explicitly disable ipv6 and this allows goss to account for that. Since we do not explicitly support ipv6 we shouldn't be strict about not having it enabled or disabled.

```release-note

```


goss/oracle/kernel-params.yaml
```diff
@@ -7,7 +7,7 @@ kernel-param:
     value: "1"
 
   net.ipv6.conf.all.disable_ipv6:
```
dkoshkin, line 9: If it can be either 0 or 1, should we just remove this check?

faiq: i thought of that as well, but i thought this would be a less big change. lets just remove it instead

dkoshkin (approved):

PR #1243, opened 2024-12-20
fix: fail on curl errors in devkit build

dkoshkin:
Previously 4xx and 5xx errors were ignored, which resulted in an invalid devkit image. This then lead to situations where the image was never rebuilt since the same image with that SHA was already in Dockerhub.

```release-note

```

### DIFF
--- a/Dockerfile.devkit
+++ b/Dockerfile.devkit
@@ -72,40 +72,46 @@ RUN mkdir -p /usr/share/ansible/collections \
     -p /usr/share/ansible/collections
 
 # hadolint ignore=DL4006
-RUN  curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz |tar xzf - -C /usr/local/bin
+RUN curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz | tar xzf - -C /usr/local/bin
+
+# Goss is used for testing the image after it is built.
+# The binary will be copied to the remote host and only linux-amd64 is supported.
 ARG GOSS_VERSION=v0.3.23
-RUN curl -L "https://github.com/goss-org/goss/releases/download/${GOSS_VERSION}/goss-linux-amd64" -o /usr/local/bin/goss-amd64
-RUN chmod +rx /usr/local/bin/goss-amd64
-ARG BUILDARCH
-RUN ln -s /usr/local/bin/goss-${BUILDARCH} /usr/local/bin/goss
-RUN curl -o /opt/amazon-ssm-agent.rpm https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm
+RUN \
+    curl --fail -v -L -o /usr/local/bin/goss-amd64 "https://github.com/goss-org/goss/releases/download/${GOSS_VERSION}/goss-linux-amd64" \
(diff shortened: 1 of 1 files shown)


Dockerfile.devkit
```diff
@@ -72,40 +72,46 @@ RUN mkdir -p /usr/share/ansible/collections \
...
+  curl --fail -v -o /opt/kubectl-${KUBERNETES_VERSION}-0-fips.rpm https://packages.d2iq.com/konvoy/stable/linux/repos/el/kubernetes-v${KUBERNETES_VERSION}-fips/x86_64/kubectl-${KUBERNETES_VERSION}-0.x86_64.rpm && \
+  curl --fail -v -o /opt/kubeadm-${KUBERNETES_VERSION}-0-fips.rpm https://packages.d2iq.com/konvoy/stable/linux/repos/el/kubernetes-v${KUBERNETES_VERSION}-fips/x86_64/kubeadm-${KUBERNETES_VERSION}-0.x86_64.rpm && \
+  curl --fail -v -o /opt/kubelet-${KUBERNETES_VERSION}-0-fips.rpm https://packages.d2iq.com/konvoy/stable/linux/repos/el/kubernetes-v${KUBERNETES_VERSION}-fips/x86_64/kubelet-${KUBERNETES_VERSION}-0.x86_64.rpm && \
   export CRICTL_TOOLS_VERSION="$(echo ${KUBERNETES_VERSION} | cut -d. -f1-2).1" && \
-  curl -o  /opt/cri-tools-${CRICTL_TOOLS_VERSION}-0-fips.rpm https://packages.d2iq.com/konvoy/stable/linux/repos/el/kubernetes-v${KUBERNETES_VERSION}-nokmem/x86_64/cri-tools-${CRICTL_TOOLS_VERSION}-0.x86_64.rpm && \
+  curl --fail -v -o /opt/cri-tools-${CRICTL_TOOLS_VERSION}-0-fips.rpm https://packages.d2iq.com/konvoy/stable/linux/repos/el/kubernetes-v${KUBERNETES_VERSION}-fips/x86_64/cri-tools-${CRICTL_TOOLS_VERSION}-0.x86_64.rpm && \
```
dkoshkin, line 110: Previously it was was pulling the nokmem version for cri-tools. Those binaries don't get modified for FIPS and are equivalent,  but still better to pull the correct one.

Dockerfile.devkit
```diff
@@ -72,40 +72,46 @@ RUN mkdir -p /usr/share/ansible/collections \
...
 # hadolint ignore=DL4006
-RUN  curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz |tar xzf - -C /usr/local/bin
+RUN curl -Lf https://github.com/mesosphere/mindthegap/releases/download/v"${MINDTHEGAP_VERSION}"/mindthegap_v"${MINDTHEGAP_VERSION}"_linux_amd64.tar.gz | tar xzf - -C /usr/local/bin
+
+# Goss is used for testing the image after it is built.
+# The binary will be copied to the remote host and only linux-amd64 is supported.
 ARG GOSS_VERSION=v0.3.23
-RUN curl -L "https://github.com/goss-org/goss/releases/download/${GOSS_VERSION}/goss-linux-amd64" -o /usr/local/bin/goss-amd64
-RUN chmod +rx /usr/local/bin/goss-amd64
-ARG BUILDARCH
```
dkoshkin, line 79: We were inconsistently using `BUILDARCH` here, this goss binary is used on the remove machine where only linux-amd64 is supported.
The Dockerfile will always look for the amd64 version
https://github.com/mesosphere/konvoy-image-builder/blob/187d011f8f9871fb0738821f2807c664e3653961/Dockerfile#L32-L33

supershal: thanks for fixing this.

supershal (approved):
Thank you.

PR #1255, opened 2025-01-30
feat: adds custom datasource for ubuntu 22.04

faiq:

```release-note

```


ansible/roles/providers/tasks/misc.yml
```diff
@@ -101,6 +101,24 @@
     mode: 0644
   when: ansible_os_family not in ["Debian", "Flatcar"] and system_cloud_init_version is version('20.0.0', '>') and system_cloud_init_version is version('23.3.0', '<')
 
+- name: Create cloud-init custom data source list
```
supershal, line 104: why do we need to copy this again?

faiq: looks like this was an oversight

supershal (approved):
Thank. you

PR #1262, opened 2025-02-21
fix: moves sysctl settings to a common place for preprovisioned and providers

faiq:
Sets sysctl settings for inotify in a common place between preprovisioned and cloud providers. 

*[NCN-105830](https://jira.nutanix.com/browse/NCN-105830)

```release-note
sets sysctl settings for inotify on preprovisioned nodes
```

### DIFF
--- a/ansible/roles/config/tasks/main.yaml
+++ b/ansible/roles/config/tasks/main.yaml
@@ -34,3 +34,15 @@
     dest: "/etc/containerd/config.toml"
   notify:
     - restart containerd
+
+- name: set and persist kernel params
+  ansible.posix.sysctl:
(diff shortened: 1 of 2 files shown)


dkoshkin (approved):

thunderboltsid (approved):

supershal (approved):
Previously this task was not running for preprovisioned usecase since it will only run if `packer_builder_type` is set. this variable is not set for preprovisioned builders. https://github.com/mesosphere/konvoy-image-builder/blob/main/ansible/roles/providers/tasks/main.yml#L32

Setting it here will always set this setting for all the providers.

PR #1267, opened 2025-03-12
feat: bump the Kubernetes version to v1.32.3

rajgitnutanix:
 Bump to latest Kubernetes release in KIB

(https://jira.nutanix.com/browse/NCN-105350)

### DIFF
--- a/ansible/group_vars/all/defaults.yaml
+++ b/ansible/group_vars/all/defaults.yaml
@@ -5,23 +5,23 @@ python_path: ""
 # If it's not there, the kubernetes_full_version will have "None" for a version number.
 #
 # IMPORTANT When you update kubernetes_version, also update crictl_version.
-kubernetes_version: "1.31.4"
+kubernetes_version: "1.32.2"
 kubernetes_major_minor: "{{ (kubernetes_version.split('.') | map('trim'))[:2] | join('.') }}"
 kubernetes_semver: "v{{ kubernetes_version }}"
 
-kubernetes_cni_version: "1.5.1"
+kubernetes_cni_version: "1.6.0"
 
 # The crictl CLI is released as part of the http://sigs.k8s.io/cri-tools project.
 # The project release closely follows the Kubernetes release cycle, and uses a
 # nearly identical version scheme.
 # IMPORTANT When you update crictl_version, also update crictl_sha256.
-crictl_version: "{{ kubernetes_major_minor }}.1"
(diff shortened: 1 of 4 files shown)


ansible/group_vars/all/defaults.yaml
```diff
@@ -5,7 +5,7 @@ python_path: ""
 # If it's not there, the kubernetes_full_version will have "None" for a version number.
 #
 # IMPORTANT When you update kubernetes_version, also update crictl_version.
-kubernetes_version: "1.31.4"
+kubernetes_version: "1.32.2"
```
dkoshkin, line 8: On line 12 here `kubernetes_cni_version` needs to match what is built in the package.

https://github.com/mesosphere/dkp-release/pull/157/files#diff-facf10ac7a5b9c0da822997252ce74ce0e278ec1d2863d3f061b51272cb8155aR45

dkoshkin (changes requested):

ansible/group_vars/all/defaults.yaml
rajgitnutanix: ok will do as suggested

ansible/group_vars/all/defaults.yaml
```diff
@@ -5,11 +5,11 @@ python_path: ""
...
```
dkoshkin, line 8: crictl version is now 1.32.0, change on line 18.

https://github.com/mesosphere/dkp-release/actions/runs/13629790801/job/38095688608#step:6:9689

rajgitnutanix: ok

dkoshkin:
If you are trying to fix the linter, please stop running e2e tests and remove the label, they are expensive.

dkoshkin (approved):

rajgitnutanix:
> If you are trying to fix the linter, please stop running e2e tests and remove the label, they are expensive.
> 
> If this is happening on `main` too please open a separate PR against main instead of including the change here.

Noted and  I  worked  on it as suggested .

dkoshkin (approved):
Thanks, looks good! The rest of the e2e tests are either flaky or in the case of Rocky 9.1 are deprecated.

rajgitnutanix:
> Thanks, looks good! The rest of the e2e tests are either flaky or in the case of Rocky 9.1 are deprecated.

Thank you @dkoshkin for the confirmation

PR #1272, opened 2025-03-20
fix: custom cloudinit script

dkoshkin:
This fixes the linter error. While looking at this, I also noticed that the second log message was using the wrong variable.

```release-note

```


supershal (approved):
Thanks for fixing this. it will make the linter happy.

dkoshkin:
Kicked off an e2e test with Ubuntu 22.04 AMI https://github.com/mesosphere/konvoy2/actions/runs/13979854864/job/39142782231

PR #1277, opened 2025-03-20
fix: simplifies rhel deps

faiq:

```release-note

```


ansible/group_vars/all/system.yaml
```diff
@@ -7,11 +7,6 @@ kubernetes_deb_repository_url: "https://pkgs.k8s.io/core:/stable:/v{{ kubernetes
 kubernetes_deb_gpg_key_url: "https://pkgs.k8s.io/core:/stable:/v{{ kubernetes_major_minor }}/deb/Release.key"
 kubernetes_deb_release_name: "/"
 
-# containerd package
-# Appstream is enabled by default in rhel8, so install the package from local repositories in that case
-docker_rpm_container_selinux_package_url: "{{ 'http://mirror.centos.org/centos/7/extras/x86_64/Packages/container-selinux-2.107-3.el7.noarch.rpm' if ansible_distribution_major_version|int < 8 else 'container-selinux' }}"
```
faiq, line 12: this stuff was used in centos 7.

bundles/redhat8.10/packages.txt.gotmpl
```diff
@@ -2,14 +2,14 @@ audit
...
 chrony
-ebtables
 open-vm-tools
 python2-pip
 python3-pip
+python3-requests
+python3-netifaces
 socat
 sysstat
-yum-utils
```
faiq, line 11: i don't think we need these here.

thunderboltsid (approved):

ansible/roles/containerd/tasks/redhat.yaml
```diff
@@ -39,15 +39,13 @@
   delay: 6
   with_items:
     - tar
-    - container-selinux
```
supershal, line 42: why this removed here?
`install container-selinux rpm package` task install container-selinux for all the OS but only if offline_mode_enabled?

ansible/roles/containerd/tasks/redhat.yaml
```diff
@@ -39,15 +39,13 @@
...
   when:
     - ansible_distribution == 'OracleLinux'
 
 - name: install container-selinux rpm package
   yum:
     name: "{{
       'container-selinux' if offline_mode_enabled
```
supershal, line 48: should remove the if condition here? The `When` condition is already taking care of it.

ansible/roles/packages/tasks/redhat.yaml
```diff
@@ -24,8 +24,8 @@
       - "{{ 'iptables-' if ansible_distribution_major_version == '8' }}ebtables"
       - open-vm-tools
       - python3-pip
-      - "python{{ '3' if ansible_distribution_major_version == '8' }}-netifaces"
```
supershal, line 27: good riddance. :)

bundles/redhat8.10/packages.txt.gotmpl
```diff
@@ -21,21 +21,17 @@ kubeadm-{{ .KubernetesVersion }}-0
...
 iptables
 iscsi-initiator-utils
-socat
 gcc
 glibc-devel
 glibc-gconv-extra
 make
 libseccomp
 iproute-tc
-gssproxy
```
supershal, line 36: I think these were added for Nutanix CSI. I didnt wanted to two versions of the package list so I added it here too.

faiq: let me double check the requirements

ansible/roles/packages/tasks/redhat.yaml
faiq: sayonara!

ansible/roles/containerd/tasks/redhat.yaml
faiq: this task only should install packages for OracleLinux -- in this case, it's just `tar` that it needs. container-selinux should be downloaded for all RHEL like operating systems and that happens on the next task.

at some point, we added this logic and after a few changes we got this very confusing result.

bundles/redhat8.10/packages.txt.gotmpl
faiq: these are required 

```   - iscsi-initiator-utils
      - nfs-utils
      - lvm2
      - xfsprogs
      - ipvsadm
      - lsscsi
      - sysstat
```
so we should be OK

PR #1311, opened 2025-06-11
fix: adds download name to file for containerd download

faiq:

adds the containerd file so we can see what version we are attempting to download. this will help us validate if we're supporting rocky 9.6 for azure

```release-note

```

### DIFF
--- a/ansible/roles/containerd/tasks/install.yaml
+++ b/ansible/roles/containerd/tasks/install.yaml
@@ -15,7 +15,7 @@
   when:
     - not containerd_tar_file_exists.stat.exists
 
-- name: download containerd
+- name: download containerd for {{ containerd_tar_file }}
   get_url:
     url: "{{ containerd_base_url }}/{{ containerd_tar_file }}"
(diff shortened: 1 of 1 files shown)


ansible/roles/containerd/tasks/install.yaml
```diff
@@ -15,7 +15,7 @@
   when:
     - not containerd_tar_file_exists.stat.exists
 
-- name: download containerd
+- name: download containerd for {{ containerd_tar_file }}
```
dkoshkin, line 18: nit: is this name better?
```suggestion
- name: download containerd {{ containerd_tar_file }}
```

Also do we still want this, will it be displayed correctly when air-gapped with `containerd_tar_file_exists.stat.exists=true`?

dkoshkin (approved):
Thanks

ansible/roles/containerd/tasks/install.yaml
faiq: i think the "for" in this case adds a bit more context so i prefer to keep it

